PR #25, opened 2021-07-15
Workaround CUDA 11.0 ICE on centos7

robertmaynard:


nvbench/detail/statistics.cuh
```diff
@@ -24,6 +24,7 @@
 #include <limits>
 #include <numeric>
 #include <vector>
+#include <cmath>
```
allisonvacanti, line 27: Alphabetize these

allisonvacanti (changes requested):
LGTM aside from a minor style issue

allisonvacanti (approved):

PR #26, opened 2021-07-21
Compile as many sources as C++ to improve build times

robertmaynard:
This takes compilation time on my machine from 12.13sec to 9.47 in parallel or 93sec to 55sec in serial.


allisonvacanti (approved):
LGTM 👍

PR #27, opened 2021-08-17
Tag unused variables

cliffburdick:
When compiling with `-Wall -Werror` unused variables in a couple files were giving errors.

### DIFF
--- a/nvbench/printer_base.cuh
+++ b/nvbench/printer_base.cuh
@@ -121,10 +121,10 @@ private:
   virtual void do_print_device_info() {}
   virtual void do_print_log_preamble() {}
   virtual void do_print_log_epilogue() {}
-  virtual void do_log(nvbench::log_level, const std::string &) {}
-  virtual void do_log_run_state(const nvbench::state &) {}
-  virtual void do_print_benchmark_list(const benchmark_vector &) {}
-  virtual void do_print_benchmark_results(const benchmark_vector &benches) {}
+  virtual void do_log([[maybe_unused]] nvbench::log_level, const std::string &) {}
+  virtual void do_log_run_state([[maybe_unused]] const nvbench::state &) {}
+  virtual void do_print_benchmark_list([[maybe_unused]] const benchmark_vector &) {}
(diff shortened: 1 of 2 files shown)


nvbench/printer_base.cuh
```diff
@@ -121,10 +121,10 @@ private:
...
   virtual void do_print_log_preamble() {}
   virtual void do_print_log_epilogue() {}
-  virtual void do_log(nvbench::log_level, const std::string &) {}
-  virtual void do_log_run_state(const nvbench::state &) {}
-  virtual void do_print_benchmark_list(const benchmark_vector &) {}
-  virtual void do_print_benchmark_results(const benchmark_vector &benches) {}
+  virtual void do_log([[maybe_unused]] nvbench::log_level, const std::string &) {}
+  virtual void do_log_run_state([[maybe_unused]] const nvbench::state &) {}
+  virtual void do_print_benchmark_list([[maybe_unused]] const benchmark_vector &) {}
+  virtual void do_print_benchmark_results([[maybe_unused]] const benchmark_vector &benches) {}
```
allisonvacanti, line 127: Rather than add `[[maybe_unused]]`, let's just remove the identifier name here, eg.

```
virtual void do_print_benchmark_results(const benchmark_vector &) {}
```

nvbench/string_axis.cuh
```diff
@@ -55,7 +55,7 @@ private:
   {
     return m_values[i];
   }
-  std::string do_get_description(std::size_t i) const final { return {}; }
+  std::string do_get_description([[maybe_unused]] std::size_t i) const final { return {}; }
```
allisonvacanti, line 58: Same as above, let's go with:

```
std::string do_get_description(std::size_t) const final { return {}; }
```

allisonvacanti (changes requested):
Can you make the suggested changes? They should have the same effect and are more consistent with how this is handled elsewhere in the library.

allisonvacanti:
LGTM, thanks!

allisonvacanti (approved):

Issue #5, opened 2021-03-18
[FEA] Built-in bandwidth measurement via CUPTI


PointKernel:
Does it make sense to introduce profiling tools like CUPTI into nvbench?

In that way, nvbench can provide precise bandwidth measurement by using CUPTI metrics like `gld_throughput` and `gst_throughput`.

Instead of manually specifying data type and size:
```
state.add_global_memory_reads<InputType>(size);
state.add_global_memory_writes<OutputType>(size);
```
Users may simply use `NVBENCH_BENCH(bm).add_global_memory_bandwidth();` to add such measurement into their tests.

allisonvacanti:
This is definitely something I want to have at some point. The current plan is to eventually add both CUPTI and NVML as optional dependencies and use them to implement these sorts of conveniences.

PointKernel:
> This is definitely something I want to have at some point. The current plan is to eventually add both CUPTI and NVML as optional dependencies and use them to implement these sorts of conveniences.

That will be great!

PR #29, opened 2021-09-04
CUPTI support

senior-zero:
Initial CUPTI support enables collection of the following metrics:
- L1 hit rates
- L2 hit rates
- DRAM throughput
- Store efficiency
- Load efficiency

This addresses [the following issue](https://github.com/NVIDIA/nvbench/issues/5). I haven't enabled CUPTI by default because there'll be a crash on Windows systems if the CUPTI lib64 directory isn't in PATH. To enable CUPTI support, it's sufficient to add `-DNVBench_ENABLE_CUPTI=ON`. After that, it's possible to specify that any of the following metrics are required:

```cuda
state.collect_dram_throughput();
state.collect_l1_hit_rates();
state.collect_l2_hit_rates();
state.collect_loads_efficiency();
state.collect_stores_efficiency();
```

I've added an example `examples/auto_throughput.cu` which produces the following output on both Linux and Windows:

| T | Stride | Elements | HBWPeak | LoadEff | StoreEff | L1HitRate | L2HitRate | Samples |  CPU Time  | Noise |  GPU Time  | Noise |  Elem/s  | Batch GPU  | Batch |
|---|--------|----------|---------|---------|----------|-----------|-----------|---------|------------|-------|------------|-------|----------|------------|-------|
| 1 |      1 | 33554432 |  91.35% | 100.00% |  100.00% |     0.00% |    50.02% |     11x | 359.400 us | 0.80% | 317.719 us | 0.15% | 105.610G | 318.293 us |   12x |
| 1 |      4 | 33554432 |  94.87% |  25.00% |  100.00% |     0.00% |    20.03% |     11x | 814.845 us | 0.34% | 773.027 us | 0.04% |  43.407G | 831.232 us |   12x |
| 2 |      1 | 33554432 |  85.71% |  50.00% |  100.00% |    30.17% |    51.88% |     11x | 737.191 us | 2.48% | 694.644 us | 2.72% |  48.304G | 747.861 us |   12x |
| 2 |      4 | 33554432 |  75.09% |  12.50% |  100.00% |    35.65% |    28.64% |     11x |   2.060 ms | 1.36% |   2.018 ms | 1.38% |  16.630G |   2.170 ms |   12x |

### DIFF
--- a/nvbench/CMakeLists.txt
+++ b/nvbench/CMakeLists.txt
@@ -26,6 +26,10 @@ set(srcs
   detail/state_generator.cxx
 )
 
+if (NVBench_ENABLE_CUPTI)
+  list(APPEND srcs detail/measure_cupti.cu cupti_profiler.cxx)
+endif()
+
 if (NVBench_ENABLE_NVML)
   list(APPEND srcs internal/nvml.cxx)
 endif()
--- a/nvbench/cupti_profiler.cuh
+++ b/nvbench/cupti_profiler.cuh
@@ -1,0 +1,129 @@
+/*
+ *  Copyright 2021 NVIDIA Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 with the LLVM exception
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.
+ *
+ *  You may obtain a copy of the License at
+ *
+ *      http://llvm.org/foundation/relicensing/LICENSE.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#pragma once
+
+#include <nvbench/config.cuh>
+#include <nvbench/device_info.cuh>
+
+#include <string>
+#include <vector>
+#include <optional>
+
+namespace nvbench::detail
+{
+
+
+#ifdef NVBENCH_HAS_CUPTI
+/**
+ * Pass required metrics in the constructor and organize your code as follows
+ * to get counters back:
+ *
+ * ```
+ * cupti_profiler cupti(
+ *      nvbench::device_info{0},
+ *      { "l1tex__t_sector_hit_rate.pct" });
+ *
+ * cupti->prepare_user_loop();
+ *
+ * do
+ * {
+ *   m_measure.m_cupti->start_user_loop();
+ *
+ *   kernel_1<<<1, 1>>>();
+ *   // ...
+ *   kernel_n<<<1, 1>>>();
+ *
+ *   m_measure.m_cupti->stop_user_loop();
+ * } while(cupti->is_replay_required());
+ *
+ * cupti->process_user_loop();
+ *
+ * auto result = m_cupti->get_counter_values();
+ * ```
+ *
+ * Check Perfworks Metric table here for the full list of metrics:
+ * https://docs.nvidia.com/cupti/r_main.html#metrics-reference-7x
+ */
+class cupti_profiler
+{
+  bool m_available {};
+  std::string m_chip_name;
+
+  // Counter data
+  std::vector<std::string> m_metric_names;
+  std::vector<std::uint8_t> m_data_image_prefix;
+  std::vector<std::uint8_t> m_config_image;
+  std::vector<std::uint8_t> m_data_image;
+  std::vector<std::uint8_t> m_data_scratch_buffer;
+  std::vector<std::uint8_t> m_availability_image;
+  nvbench::device_info m_device;
+
+  // CUPTI runs a series of replay passes, where each pass contains a sequence
+  // of ranges. Every metric enabled in the configuration is collected
+  // separately per unique range in the pass. CUPTI supports auto and
+  // user-defined ranges. With auto range mode, ranges are defined around each
+  // kernel automatically. In the user range mode, ranges are defined manually.
+  // We define a single user range for the whole measurement.
+  static const int m_num_ranges = 1;
+
+public:
+  // Move only
+  cupti_profiler(cupti_profiler &&) noexcept;
+  cupti_profiler &operator=(cupti_profiler &&) noexcept;
+
+  cupti_profiler(const cupti_profiler &) = delete;
+  cupti_profiler &operator=(const cupti_profiler &) = delete;
+
+  cupti_profiler(nvbench::device_info device,
+                 std::vector<std::string> &&metric_names);
+  ~cupti_profiler();
+
+  [[nodiscard]] bool is_initialized() const;
+
+  /// Should be called before replay loop
+  void prepare_user_loop();
+
+  /// Should be called before any kernel calls in the replay loop
+  void start_user_loop();
+
+  /// Should be called after all kernel calls in the replay loop
+  void stop_user_loop();
+
+  /// Should be called after the replay loop
+  void process_user_loop();
+
+  /// Indicates whether another iteration of the replay loop is required
+  [[nodiscard]] bool is_replay_required();
+
+  /// Returns counters for metrics requested in the constructor
+  [[nodiscard]] std::vector<double> get_counter_values();
+
+private:
+  void initialize_profiler();
+  void initialize_chip_name();
+  void initialize_availability_image();
+  static void initialize_nvpw();
+  void initialize_config_image();
+  void initialize_counter_data_prefix_image();
+  void initialize_counter_data_image();
+};
+#endif
+
+
+} // namespace nvbench::detail
--- a/nvbench/detail/measure_cupti.cu
+++ b/nvbench/detail/measure_cupti.cu
@@ -1,0 +1,263 @@
+/*
+ *  Copyright 2021 NVIDIA Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 with the LLVM exception
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.
+ *
+ *  You may obtain a copy of the License at
+ *
+ *      http://llvm.org/foundation/relicensing/LICENSE.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include <nvbench/benchmark_base.cuh>
+#include <nvbench/detail/measure_cupti.cuh>
+#include <nvbench/detail/throw.cuh>
+#include <nvbench/printer_base.cuh>
+#include <nvbench/state.cuh>
+
+#include <fmt/format.h>
+
+#include <algorithm>
+#include <stdexcept>
+
+namespace nvbench::detail
+{
+
+namespace
+{
+
+enum class metric_id : int
+{
+  dram_peak_sustained_throughput = 0,
+  global_load_efficiency,
+  global_store_efficiency,
+  l1_hit_rate,
+  l2_hit_rate,
+
+  count
+};
+
+template <metric_id id>
+struct metric_traits;
+
+template <>
+struct metric_traits<metric_id::dram_peak_sustained_throughput>
+{
+  static constexpr const char *metric_name =
+    "dram__throughput.avg.pct_of_peak_sustained_elapsed";
+
+  static constexpr const char *summary =
+    "Peak Sustained Global Memory Throughput (HW)";
+
+  static constexpr const char *hint       = "percentage";
+  static constexpr const char *short_name = "HBWPeak";
+
+  static constexpr const char *description =
+    "The utilization level of the device memory relative to the peak "
+    "utilization.";
+
+  static constexpr double divider = 100.0;
+
+  static bool is_collected(nvbench::state &m_state)
+  {
+    return m_state.is_dram_throughput_collected();
+  };
+};
+
+template <>
+struct metric_traits<metric_id::global_load_efficiency>
+{
+  static constexpr const char *metric_name =
+    "smsp__sass_average_data_bytes_per_sector_mem_global_op_ld.pct";
+
+  static constexpr const char *summary    = "Global Load Efficiency (HW)";
+  static constexpr const char *hint       = "percentage";
+  static constexpr const char *short_name = "LoadEff";
+
+  static constexpr const char *description =
+    "Ratio of requested global memory load throughput to required global "
+    "memory load throughput expressed as percentage.";
+
+  static constexpr double divider = 100.0;
+
+  static bool is_collected(nvbench::state &m_state)
+  {
+    return m_state.is_loads_efficiency_collected();
+  };
+};
+
+template <>
+struct metric_traits<metric_id::global_store_efficiency>
+{
+  static constexpr const char *metric_name =
+    "smsp__sass_average_data_bytes_per_sector_mem_global_op_st.pct";
+
+  static constexpr const char *summary    = "Global Store Efficiency (HW)";
+  static constexpr const char *hint       = "percentage";
+  static constexpr const char *short_name = "StoreEff";
+
+  static constexpr const char *description =
+    "Ratio of requested global memory store throughput to required global "
+    "memory store throughput expressed as percentage.";
+
+  static constexpr double divider = 100.0;
+
+  static bool is_collected(nvbench::state &m_state)
+  {
+    return m_state.is_stores_efficiency_collected();
+  };
+};
+
+template <>
+struct metric_traits<metric_id::l1_hit_rate>
+{
+  static constexpr const char *metric_name = "l1tex__t_sector_hit_rate.pct";
+  static constexpr const char *summary     = "L1 Cache Hit Rate (HW)";
+  static constexpr const char *hint        = "percentage";
+  static constexpr const char *short_name  = "L1HitRate";
+  static constexpr const char *description = "Hit rate at L1 cache.";
+  static constexpr double divider          = 100.0;
+
+  static bool is_collected(nvbench::state &m_state)
+  {
+    return m_state.is_l1_hit_rate_collected();
+  };
+};
+
+template <>
+struct metric_traits<metric_id::l2_hit_rate>
+{
+  static constexpr const char *metric_name = "lts__t_sector_hit_rate.pct";
+  static constexpr const char *summary     = "L2 Cache Hit Rate (HW)";
+  static constexpr const char *hint        = "percentage";
+  static constexpr const char *short_name  = "L2HitRate";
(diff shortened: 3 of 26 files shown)


nvbench/CMakeLists.txt
```diff
@@ -64,6 +69,15 @@ set_target_properties(nvbench PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY "${NVBench_LIBRARY_OUTPUT_DIR}"
   RUNTIME_OUTPUT_DIRECTORY "${NVBench_EXECUTABLE_OUTPUT_DIR}"
 )
+if (NVBench_ENABLE_CUPTI)
+  if (WIN32)
+    message("Don't forget to add ${CUDAToolkit_LIBRARY_ROOT}/extras/CUPTI/lib64 in your PATH!")
```
allisonvacanti, line 74: @robertmaynard Do you know anything about this?

@senior-zero mentioned that he had to modify `PATH` like this when linking to CUPTI on windows. Is there a way to fix this without bugging the user?

robertmaynard: Link to the `CUDA::cuda_driver` and `CUDA::cupti` targets instead of having the explicit paths in target_include_directories, and target_link_directories plus the usage of raw libraries ( cuda, cupti ) in target_link_libraries

nvbench/detail/measure_cupti.cu
```diff
@@ -0,0 +1,288 @@
...
+                  "https://developer.nvidia.com/"
+                  "nvidia-development-tools-solutions-err_nvgpuctrperm-"
+                  "permission-issue-performance-counters");
+    }
+  }
+
+  const auto device = m_state.get_device();
+  if (!device)
+  {
+    throw std::runtime_error(fmt::format("{}:{}: Device required for `cupti` "
```
allisonvacanti, line 228: You may find the `NVBENCH_THROW` macro in `nvbench/detail/throw.cuh` helpful here :)

nvbench/detail/measure_cupti.cu
```diff
@@ -0,0 +1,288 @@
...
+    throw std::runtime_error(fmt::format("{}:{}: Internal error: Current "
+                                         "device is not active.",
+                                         __FILE__,
+                                         __LINE__));
+  }
+}
+
+
+template <metric_ids id>
+void gen_summary(std::size_t result_id,
```
allisonvacanti, line 244: anon namespace, similar naming suggestion as the other metric helpers (`gen_summaries_impl`)

nvbench/detail/measure_cupti.cu
```diff
@@ -0,0 +1,288 @@
...
+
+#include <algorithm>
+#include <stdexcept>
+
+
+namespace nvbench::detail
+{
+
+
+enum metric_ids : int
```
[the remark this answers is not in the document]
allisonvacanti: Use a scoped enum for improved type-safety: `enum class metric_ids`.

We should also decide whether we want enum scope names to be plural or singular, e.g. `metric_ids` vs. `metric_id`. I think I prefer singular, but I haven't been consistent about this...

nvbench/detail/measure_cupti.cu
```diff
@@ -0,0 +1,288 @@
...
+void gen_summaries(nvbench::state &state,
+                   const std::vector<double> &result)
+{
+  gen_summary<metric_ids::dram_peak_sustained_throughput>(0, state, result);
+}
+
+
+void measure_cupti_base::generate_summaries()
+{
+  try
```
allisonvacanti, line 279: Consider the function scope try/catch syntax when the entire function is in a try block:

```
void function()
try {
  ...
}
catch (...)
{
  ...
}
```

nvbench/detail/measure_cupti.cuh
```diff
@@ -0,0 +1,160 @@
...
+ */
+
+#pragma once
+
+#include <nvbench/blocking_kernel.cuh>
+#include <nvbench/cpu_timer.cuh>
+#include <nvbench/cuda_call.cuh>
+#include <nvbench/cuda_timer.cuh>
+#include <nvbench/device_info.cuh>
+#include <nvbench/cupti_profiler.h>
```
allisonvacanti, line 26: alphabetize

nvbench/detail/measure_cupti.cuh
```diff
@@ -0,0 +1,160 @@
...
+  {
+    NVBENCH_CUDA_CALL(cudaStreamSynchronize(m_launch.get_stream()));
+  }
+
+  nvbench::state &m_state;
+
+  nvbench::launch m_launch;
+  nvbench::detail::l2flush m_l2flush;
+
+  std::optional<cupti_profiler> m_cupti;
```
allisonvacanti, line 79: Why optional?

nvbench/device_info.cuh
```diff
@@ -175,6 +183,30 @@ struct device_info
   /// @return True if ECC is enabled on this device.
   [[nodiscard]] bool get_ecc_state() const { return m_prop.ECCEnabled; }
 
+  /// @return True if CUPTI supports this device.
+  [[nodiscard]] bool is_CUPTI_supported() const
```
allisonvacanti, line 187: Some API capitalizes `CUPTI`, some just uses lowercase `cupti`. Let's just stick with lowercase throughout the API, since other abbreviations are handled that way.

nvbench/device_info.cuh
```diff
@@ -64,6 +65,13 @@ struct device_info
   void set_active() const
   {
     NVBENCH_CUDA_CALL(cudaSetDevice(m_id));
+
+#ifdef NVBench_CUPTI_AVAILABLE
```
allisonvacanti, line 69: `NVBENCH_HAS_CUPTI`

nvbench/device_info.cuh
```diff
@@ -175,6 +183,30 @@ struct device_info
   /// @return True if ECC is enabled on this device.
   [[nodiscard]] bool get_ecc_state() const { return m_prop.ECCEnabled; }
 
+  /// @return True if CUPTI supports this device.
+  [[nodiscard]] bool is_CUPTI_supported() const
+  {
+#ifdef NVBench_CUPTI_AVAILABLE
```
allisonvacanti, line 189: `NVBENCH_HAS_CUPTI`

nvbench/device_info.cuh
```diff
@@ -175,6 +183,30 @@ struct device_info
...
+    return m_prop.major >= 7;
+#else
+    return false;
+#endif
+  }
+
+#ifdef NVBench_CUPTI_AVAILABLE
```
allisonvacanti, line 196: `NVBENCH_HAS_CUPTI`

nvbench/main.cuh
```diff
@@ -45,9 +45,16 @@
     return 1;                                                                  \
   }
 
+#ifdef NVBench_CUPTI_AVAILABLE
```
allisonvacanti, line 48: `NVBENCH_HAS_CUPTI`

nvbench/detail/state_exec.cuh
```diff
@@ -74,16 +75,37 @@ void state::exec(ExecTags tags, KernelLauncher &&kernel_launcher)
     constexpr bool use_blocking_kernel = !(tags & no_block);
     if constexpr (tags & timer)
     {
+      // Estimate bandwidth here
+      #ifdef NVBench_CUPTI_AVAILABLE
+      if (is_cupti_required())
```
allisonvacanti, line 80: Use explicit `this->` for member function calls.

nvbench/detail/state_exec.cuh
```diff
@@ -74,16 +75,37 @@ void state::exec(ExecTags tags, KernelLauncher &&kernel_launcher)
...
       measure();
     }
     else
     { // Need to wrap the kernel launcher with a timer wrapper:
       using wrapper_t = nvbench::detail::kernel_launch_timer_wrapper<KL>;
+      wrapper_t wrapper{kernel_launcher};
+
+      // Estimate bandwidth here
+      #ifdef NVBench_CUPTI_AVAILABLE
+      if (is_cupti_required())
```
allisonvacanti, line 99: `this->is_cupti_required()`

nvbench/CMakeLists.txt
```diff
@@ -24,7 +24,12 @@ set(srcs
   detail/measure_cold.cu
   detail/measure_hot.cu
   detail/state_generator.cxx
-)
+
+  detail/measure_cupti.cuh
```
allisonvacanti, line 28: Rather than unconditionally compiling these and slicing bits of them out with `NVBENCH_HAS_CUPTI`, let's do this:

```
if (NVBench_ENABLE_CUPTI)
  list(APPEND srcs [cupti files here])
endif()
```

and then remove the `NVBENCH_HAS_CUPTI` checks from these files. This clean separation will reduce the complexity of maintaining the optional dependency, since the only significant code divergence will be in `state_exec.cuh`.

We'll need to move error logging for the "cupti is not enabled but requested" case in `state_exec.cuh` with this approach.

allisonvacanti: Don't list the header files here -- it's harmless, but currently we're only listing translation units in the `srcs` list.

nvbench/CMakeLists.txt
```diff
@@ -24,7 +24,12 @@ set(srcs
...
+  detail/measure_cupti.cu
+
+  cupti_profiler.cpp
```
allisonvacanti, line 31: Use the `.cxx` extension for consistency.

nvbench/CMakeLists.txt
```diff
@@ -64,6 +69,15 @@ set_target_properties(nvbench PROPERTIES
   LIBRARY_OUTPUT_DIRECTORY "${NVBench_LIBRARY_OUTPUT_DIR}"
   RUNTIME_OUTPUT_DIRECTORY "${NVBench_EXECUTABLE_OUTPUT_DIR}"
 )
+if (NVBench_ENABLE_CUPTI)
+  if (WIN32)
+    message("Don't forget to add ${CUDAToolkit_LIBRARY_ROOT}/extras/CUPTI/lib64 in your PATH!")
+  endif ()
+  target_include_directories(nvbench PRIVATE ${CUDAToolkit_LIBRARY_ROOT}/extras/CUPTI/include)
+  target_link_directories(nvbench PUBLIC ${CUDAToolkit_LIBRARY_ROOT}/extras/CUPTI/lib64)
+  target_link_libraries(nvbench PUBLIC cuda cupti nvperf_host nvperf_target)
```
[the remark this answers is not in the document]
allisonvacanti: This may also fix the windows PATH issue above.

nvbench/cupti_profiler.cpp
```diff
@@ -0,0 +1,790 @@
...
+  initialize_chip_name();
+  initialize_availability_image();
+  initialize_nvpw();
+  initialize_config_image();
+  initialize_counter_data_prefix_image();
+  initialize_counter_data_image();
+
+  m_available = true;
+}
+catch (...)
```
allisonvacanti, line 81: Why throw away exceptions here?

nvbench/detail/measure_cupti.cu
```diff
@@ -0,0 +1,288 @@
...
+
+      if (!m_cupti->initialized())
+      {
+        m_cupti.reset();
+      }
+    }
+  }
+  catch (...)
+  {
+    m_cupti.reset();
```
allisonvacanti, line 201: Should this be logged?

nvbench/detail/measure_cupti.cu
```diff
@@ -0,0 +1,288 @@
...
+}
+
+
+void measure_cupti_base::generate_summaries()
+{
+  try
+  {
+    gen_summaries(m_state, m_cupti->get_counter_values());
+  }
+  catch (...)
```
allisonvacanti, line 283: Do we really want to throw away exceptions here?

allisonvacanti (changes requested):
This is looking great! Really nice addition -- most of my comments are minor and repetitive ;)

nvbench/CMakeLists.txt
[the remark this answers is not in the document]
allisonvacanti: Looks like Rob confirmed this while I was finishing the review 😁

nvbench/CMakeLists.txt
allisonvacanti: #46 adds NVML support and fixes a lot of the CMake issues in this PR:

- Fixes/WARs the dll PATH issues without requiring users to modify PATH
- Adds an auto-generated `nvbench/config.cuh` header instead of relying on compiler definitions.

It should be fairly easy to extend the approaches used in that PR here. I'd suggest rebasing this on #46 so we can reuse the cmake machinery.

nvbench/cupti_profiler.cpp
```diff
@@ -0,0 +1,790 @@
...
+    CUpti_Profiler_Initialize_Params_STRUCT_SIZE};
+  cupti_call(cuptiProfilerInitialize(&params));
+}
+
+
+void cupti_profiler::initialize_chip_name()
+{
+  CUpti_Device_GetChipName_Params params = {
+    CUpti_Device_GetChipName_Params_STRUCT_SIZE};
+  params.deviceIndex = device->get_id();
```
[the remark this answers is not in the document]
senior-zero: Judging on CUPTI samples, it's using CUDA ids.

nvbench/cupti_profiler.cpp
```diff
@@ -0,0 +1,790 @@
...
+  {
+    CUpti_Profiler_DeInitialize_Params params = {
+      CUpti_Profiler_DeInitialize_Params_STRUCT_SIZE};
+
+    cuptiProfilerDeInitialize(&params);
+  }
+}
+
+
+[[nodiscard]] bool cupti_profiler::initialized() const
```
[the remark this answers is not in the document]
senior-zero: I'm not sure I understand. Which case are you referring to? https://godbolt.org/z/rd8zfcMKW 
Neither bool nor std::vector has `[[nodiscard]]` in their declarations. Having `[[nodiscard]]` in the `cupti_profiler` wouldn't affect it's methods.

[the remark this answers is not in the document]
allisonvacanti: The `[[nodiscard]]` attribute is already used inside the definition of `cupti_profiler` on the declaration of the `cupti_profiler::initialize` method:

```
struct cupti_profiler
{
  // ...
  [[nodiscard]] bool initialized() const;
  // ...
};
```

and it's not necessary to repeat it here on the method definition. Ex: https://godbolt.org/z/KMT8hhW6v

I prefer to just have these attributes inside the class definition to avoid having to keep both decl/def sync'd, but it's admittedly an extremely minor nitpick 🙂

nvbench/detail/measure_cupti.cu
allisonvacanti: For reference, here's how to print a warning to the log: https://github.com/NVIDIA/nvbench/blob/main/nvbench/detail/measure_cold.cu#L158-L196

nvbench/detail/state_exec.cuh
```diff
@@ -23,10 +23,12 @@
        "Do not include it directly."
 #endif // NVBENCH_STATE_EXEC_GUARD
 
+#include <nvbench/config.cuh>
 #include <nvbench/exec_tag.cuh>
 #include <nvbench/state.cuh>
 
 #include <nvbench/detail/kernel_launcher_timer_wrapper.cuh>
+#include <nvbench/detail/measure_cupti.cuh>
```
allisonvacanti, line 31: When I run `nvbench.example.auto_throughput` with CUPTI disabled, I get

```
bin/nvbench.example.auto_throughput: symbol lookup error: bin/nvbench.example.auto_throughput: undefined symbol: nvbench::detail::measure_cupti_base::generate_summaries()
```

Wrapping this header in `#ifdef NVBENCH_HAS_CUPTI / #endif` fixes it.

nvbench/cupti_profiler.cuh
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Pass required metrics in the constructor and organize your code as follows
+ * to get counters back:
+ *
+ * ```
+ * cupti_profiler cupti(
+ *      nvbench::device_info{0},
+ *      { "l1tex__t_sector_hit_rate.pct" });
+ *
+ * cupti->prepare_to_user_loop();
```
allisonvacanti, line 42: This example is very helpful!

Small language nitpick about this method name, `prepare_user_loop` sounds more natural.

nvbench/detail/measure_cupti.cuh
```diff
@@ -0,0 +1,158 @@
...
+
+struct measure_cupti_base::kernel_launch_timer
+{
+  explicit kernel_launch_timer(measure_cupti_base &measure)
+      : m_measure{measure}
+  {}
+
+  __forceinline__ void start()
+  {
+#ifdef NVBENCH_HAS_CUPTI
```
allisonvacanti, line 92: This is redundant with line 23.

I think we can probably remove all of these checks and just ifdef out the `#include <nvbench/detail/measure_cupti.cuh>` in `state_exec.cuh`.

nvbench/detail/measure_cupti.cuh
```diff
@@ -0,0 +1,158 @@
...
+    if (m_measure.m_cupti.is_initialized())
+    {
+      m_measure.m_cupti.start_user_loop();
+    }
+#endif
+  }
+
+  __forceinline__ void stop()
+  {
+#ifdef NVBENCH_HAS_CUPTI
```
allisonvacanti, line 105: Same note as above.

nvbench/detail/measure_cupti.cuh
```diff
@@ -0,0 +1,158 @@
...
+    this->check();
+    this->run();
+    this->generate_summaries();
+  }
+
+private:
+  // Run the kernel as many times as CUPTI requires.
+  void run()
+  {
+#ifdef NVBENCH_HAS_CUPTI
```
allisonvacanti, line 138: Same as above.

nvbench/device_info.cuh
```diff
@@ -175,6 +208,30 @@ struct device_info
...
+    return false;
+#endif
+  }
+
+#ifdef NVBENCH_HAS_CUPTI
+  [[nodiscard]] CUcontext get_context() const
+  {
+    if (!is_active())
+    {
+      throw std::runtime_error("get_context is called for inactive device");
```
allisonvacanti, line 226: Move this function into `device_info.cu` (to avoid leaking fmt) and use the NVBENCH_THROW macro to embed file/line information.

nvbench/main.cuh
```diff
@@ -45,9 +46,16 @@
     return 1;                                                                  \
   }
 
+#ifdef NVBENCH_HAS_CUPTI
+#define INITIALIZE_DRIVER_API NVBENCH_DRIVER_API_CALL(cuInit(0));
```
allisonvacanti, line 50: Prefix this with `NVBENCH_` since it needs to leak into user code.

allisonvacanti (approved):
A couple minor issues in the comments, but otherwise I'm really happy with this -- it's a very useful addition and I appreciate your hard work here!

Approving, once nvml is merged and these last comments are address this is ready to go in. 👍

Issue #36, opened 2021-10-08
Option parser test failing


allisonvacanti:
```
~/nvbench/build$ ./bin/nvbench.test.option_parser
nvbench/nvbench/option_parser.cu:672: Error parsing --axis ` PO2s [ pow2 ] = 7 `:
nvbench/nvbench/option_parser.cu:694: Invalid flag for int64 axis: ` �|`
```

`git bisect` points to 15434ce9e24c38e18c92e3a61d88aac990592add, which is fairly surprising, since that really shouldn't affect behavior.

PR #37, opened 2021-10-13
Revert .cu -> .cxx change for option_parser TU.

alliepiper:
This change introduced a strange bug on GCC where a stack object is changed
seemingly randomly.

In `option_parser::update_axis`, the `flag` variable's `data()` pointer
is overwritten to point at garbage memory shortly after parsing. Nothing
related to this object is being updated when the corruption occurs.

This has been observed on gcc7+, but cannot be reproduced on MSVC.

Reverting the `option_parser.cxx` TU to be a CUDA object works
around this.

Fixes issue #36.


robertmaynard:
Ohh joy compiler bugs. Thanks for tracking this down :+1:

PR #48, opened 2021-12-17
nvbench_compare handles directories and can filter out non-interesting results 

robertmaynard:
These two features are really useful when you want to look for changes in a large collection of performance results.

### DIFF
--- a/scripts/nvbench_compare.py
+++ b/scripts/nvbench_compare.py
@@ -1,9 +1,11 @@
 #!/usr/bin/env python
 
-from colorama import Fore
+import argparse
 import json
 import math
 import sys
+
+from colorama import Fore
 
 import tabulate
 
@@ -13,22 +15,7 @@ def version_tuple(v):
 
 tabulate_version = version_tuple(tabulate.__version__)
 
-if len(sys.argv) != 3:
-    print("Usage: %s reference.json compare.json\n" % sys.argv[0])
-    sys.exit(1)
-
-with open(sys.argv[1], "r") as ref_file:
-    ref_root = json.load(ref_file)
-
-with open(sys.argv[2], "r") as cmp_file:
-    cmp_root = json.load(cmp_file)
-
-# This is blunt but works for now:
-if ref_root["devices"] != cmp_root["devices"]:
-    print("Device sections do not match.")
-    sys.exit(1)
-
-all_devices = cmp_root["devices"]
+all_devices = []
 config_count = 0
 unknown_count = 0
 failure_count = 0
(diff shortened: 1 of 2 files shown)


scripts/nvbench_compare.py
```diff
@@ -226,16 +212,23 @@ def compare_benches(ref_benches, cmp_benches):
...
                     status = Fore.RED + "FAIL" + Fore.RESET
 
-                row.append(format_duration(ref_time))
-                row.append(format_percentage(ref_noise))
-                row.append(format_duration(cmp_time))
-                row.append(format_percentage(cmp_noise))
-                row.append(format_duration(diff))
-                row.append(format_percentage(frac_diff))
-                row.append(status)
+                if abs(frac_diff) >= threshold:
```
allisonvacanti, line 215: This check could be done earlier by continuing after frac_diff is computed.

scripts/nvbench_compare.py
```diff
@@ -250,12 +243,56 @@ def compare_benches(ref_benches, cmp_benches):
...
+    help_text = "%(prog)s [reference.json compare.json | reference_dir/ compare_dir/]"
+    parser = argparse.ArgumentParser(prog='nvbench_compare', usage=help_text)
+    parser.add_argument('--threshold-diff',type=float, dest='threshold', default=0.0,
+                        help='only show benchmarks where percentage diff is >= THRESHOLD')
+
+    if len(sys.argv) < 3:
+        parser.print_help()
+        sys.exit(1)
+
+    args = parser.parse_args(sys.argv[3:])
```
allisonvacanti, line 257: Does this mean that `--threshold-diff` must be last? Would it be easy to redo this so that

```
nvbench_compare --threshold-diff 5 ref.json comp.json
```

would work too?

scripts/requirements.txt
```diff
@@ -0,0 +1,2 @@
+colorama
```
allisonvacanti, line 1: This looks redundant with the `requirements.txt` in the root of the `nvbench` repo. I'm not sure what consumes these files, but we should probably only have one.

robertmaynard: removed, I totally missed the requirements.txt in the root directory.

scripts/nvbench_compare.py
robertmaynard: Updated to allow that

scripts/nvbench_compare.py
robertmaynard: I wanted to do it after all computation in case we want to expand to have different filter modes, but for KISS I can move it up

allisonvacanti (approved):

PR #57, opened 2021-12-20
Fix UB in option parser

senior-zero:
`parse_axis_key_flag_value_spec` used `std::tie` to return references to local objects. This led to various racy runtime issues when specifying `-a` options in command line.


allisonvacanti (approved):
Yikes -- good catch!

PR #58, opened 2021-12-20
Add an `nvbench-ctl` executable.

allisonvacanti:
This will provide functionality such as clock locking (--lgm),
persistance mode (--pm), device querying (--list), version checking
(--version), and documentation (--help).

This is possible already with any nvbench executable, but having
one with a reliable name will be helpful for scripting and writing
documentation.

### DIFF
--- a/exec/nvbench-ctl.cu
+++ b/exec/nvbench-ctl.cu
@@ -1,0 +1,48 @@
+/*
+*  Copyright 2021 NVIDIA Corporation
+*
+*  Licensed under the Apache License, Version 2.0 with the LLVM exception
+*  (the "License"); you may not use this file except in compliance with
+*  the License.
+*
+*  You may obtain a copy of the License at
+*
+*      http://llvm.org/foundation/relicensing/LICENSE.txt
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*/
+
+#include <nvbench/main.cuh>
+
+#include <vector>
+
+int main(int argc, char const *const *argv)
+try
+{
+  // If no args, substitute a new argv that prints the version
+  std::vector<const char*> alt_argv;
+  if (argc == 1)
+  {
+    alt_argv.push_back("--version");
+    alt_argv.push_back(nullptr);
+    argv = alt_argv.data();
+  }
+
+  NVBENCH_MAIN_PARSE(argc, argv);
+  NVBENCH_CUDA_CALL(cudaDeviceReset());
+  return 0;
+}
+catch (std::exception & e)
+{
+  std::cerr << "\nNVBench encountered an error:\n\n" << e.what() << "\n";
+  return 1;
+}
+catch (...)
+{
+  std::cerr << "\nNVBench encountered an unknown error.\n";
+  return 1;
+}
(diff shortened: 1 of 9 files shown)


testing/cmake/test_export/CMakeLists.txt
```diff
@@ -8,11 +8,37 @@ add_executable(test_bench test_bench.cu)
...
+  set(cupti_lib_path "")
+
+  # Need to find installed libnvbench.so for installed nvbench-ctl.
+  # Not needed for build_tree test because of RUNPATH.
+  if (TEST_TYPE STREQUAL "INSTALL_TREE")
+    get_property(nvbench_config TARGET nvbench::nvbench
+      PROPERTY IMPORTED_CONFIGURATIONS
+    )
+    get_property(ctl_lib_path TARGET nvbench::nvbench
+      PROPERTY IMPORTED_LOCATION_${nvbench_config}
```
allisonvacanti, line 25: @robertmaynard I need to redo how I detect `nvbench_config`, since this code only works when there's one `IMPORTED_CONFIGURATION` for `nvbench::nvbench`. What's the proper way to determine which configuration to use?

robertmaynard: For muti-config generators you don't have a great option.

You can iterate `IMPORTED_CONFIGURATIONS` to see all the built configurations for nvbench. But from there you will need to do something like this:
```
add_test(NAME mytest
         COMMAND ${CMAKE_COMMAND} -P <script> -DCONFIG=$<CONFIG>
                            -DTEST=$<TARGET_FILE:myexe>)
```

And have `script` set the LD_LIBRARY_PATH

exec/nvbench-ctl.cu
```diff
@@ -0,0 +1,48 @@
...
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*/
+
+#include <nvbench/main.cuh>
+
+#include <vector>
+
+int main(int argc, char const *const *argv)
+try
```
karthikeyann, line 24: TIL https://en.cppreference.com/w/cpp/language/function-try-block 👍

nvbench/main.cuh
```diff
@@ -47,17 +47,22 @@
...
-#define NVBENCH_INITIALIZE_DRIVER_API NVBENCH_DRIVER_API_CALL(cuInit(0));
+#define NVBENCH_INITIALIZE_DRIVER_API NVBENCH_DRIVER_API_CALL(cuInit(0))
 #else
-#define NVBENCH_INITIALIZE_DRIVER_API
+// clang-format off
+#define NVBENCH_INITIALIZE_DRIVER_API do {} while (false)
+// clang-format on
 #endif
 
+#define NVBENCH_MAIN_PARSE(arg, argv)                                          \
```
karthikeyann, line 57: typo?
```suggestion
#define NVBENCH_MAIN_PARSE(argc, argv)                                          \
```

testing/cmake/test_export/CMakeLists.txt
allisonvacanti: Thanks -- that sounds complicated 😅

Since we're only querying this to get the libdir for `LD_LIBRARY_PATH`, I'll just emit a warning and use the first one in the list. If it breaks for multiconfig later I'll fix it properly. I don't know that we'll actually run into the combination of UNIX + multiconfig builds for this to matter.

exec/nvbench-ctl.cu
allisonvacanti: @brycelelbach showed me that a while ago -- it's handy :)

nvbench/main.cuh
allisonvacanti: Good catch! Fixed, thanks!

PR #79, opened 2022-02-11
Fix a bug in config count unit test: count number of devices as well

PointKernel:
This PR fixes a minor bug when testing `benchmark::get_config_count()`. The number of devices should be taken into consideration as well.


allisonvacanti (approved):

allisonvacanti:
Thanks!

PR #85, opened 2022-04-08
Add --disable-blocking-kernel and --profile options.

### LINKED ISSUE #45: Add `--profile` flag to fixup various profiling issues
#10 was a similar issue about adding support for profiling -- this issue resulted in the `--run-once` option, which disables warmup runs and only does a single execution of the `KernelLauncher`.

However, there is another issue when using CUDA profilers. The profiler will force the CPU to block until the kernel is executed, and this causes a deadlock when a benchmark uses the `blocking_kernel`. This can be worked around by using `nvbench::exec_tag::sync` to disable the blocking kernel, but we should add a runtime switch to disable this as well.

- [ ] Add `--disable-blocking-kernel` to simply switch off the blocking kernel
- [ ] Add `--profile`, which implies `--disable-blocking-kernel --run-once`

pauleonix:
Closes #45.
Adds a mode that forces a benchmark not to use the `blocking_kernel`, simplifying
profiling usecases. This can be enabled by any of the following methods:
- Passing `--disable-blocking-kernel` on the command line
- `NVBENCH_CREATE(...).set_no_block(true)` when declaring a benchmark
- `state.set_no_block(true)` from within the benchmark implementation.
Also adds the `--profile` flag which implies `--run-once` and `--disable-blocking-kernel`.

### DIFF
--- a/nvbench/detail/state_exec.cuh
+++ b/nvbench/detail/state_exec.cuh
@@ -64,6 +64,13 @@ void state::exec(ExecTags tags, KernelLauncher &&kernel_launcher)
     return;
   }
 
+  if (!(modifier_tags & no_block) && this->get_no_block())
+  {
+    constexpr auto no_block_tags = modifier_tags | no_block;
+    this->exec(no_block_tags, std::forward<KernelLauncher>(kernel_launcher));
+    return;
+  }
+
   // If no measurements selected, pick some defaults based on the modifiers:
   if constexpr (!measure_tags)
   {
--- a/docs/cli_help.md
+++ b/docs/cli_help.md
@@ -131,3 +131,15 @@
   * Intended for use with external profiling tools.
   * Applies to the most recent `--benchmark`, or all benchmarks if specified
     before any `--benchmark` arguments.
+
+* `--disable-blocking-kernel`
+  * Don't use the `blocking_kernel`.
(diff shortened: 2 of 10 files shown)


pauleonix:
This PR is very inspired by #32. I hope this helps.

nvbench/detail/state_exec.cuh
```diff
@@ -64,6 +64,13 @@ void state::exec(ExecTags tags, KernelLauncher &&kernel_launcher)
     return;
   }
 
+  if (!(modifier_tags & no_block) && this->get_no_block())
+  {
+    constexpr auto no_block_tags = modifier_tags | no_block;
```
allisonvacanti, line 69: This will drop the flags that specify measurement types.

```suggestion
    constexpr auto no_block_tags = tags | no_block;
```

allisonvacanti (changes requested):
Thanks for the contribution! Greatly appreciated :)

A couple of minor changes are needed, but otherwise this LGTM.

pauleonix:
Thanks for the review! The PR should me ready to be merged now.

allisonvacanti (approved):

allisonvacanti:
Looks good, merging -- thanks again!

PR #89, opened 2022-06-06
Fix a typo in README.md

PointKernel:
The proper CMake variable to set CUDA architectures should be `CMAKE_CUDA_ARCHITECTURES`.


allisonvacanti (approved):
Thanks!

PR #94, opened 2022-08-08
Update to latest version of rapids

robertmaynard:
Also ensure that we don't clobber any existing rapids.cmake file


allisonvacanti (approved):

allisonvacanti:
Thanks Rob!

PR #96, opened 2022-08-18
Fix formatting

senior-zero:
This PR changes the line limit to 100 and also uses newer policy for `SortIncludes`. I've auto formatted the source base with clang-format 14.0. The sources seems to benefit from larger line limit.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -36,7 +36,7 @@ BreakBeforeBinaryOperators: None
 BreakBeforeTernaryOperators: true
 BreakConstructorInitializers: BeforeComma
 BreakInheritanceList: BeforeComma
-ColumnLimit: 80
+ColumnLimit: 100
 CompactNamespaces: false
 ContinuationIndentWidth: 2
 IndentCaseLabels: true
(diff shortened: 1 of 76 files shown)


allisonvacanti (approved):
I'm okay with this change. It does look much nicer overall.

Before (or just after) merging this branch, add a file like [this one](https://github.com/NVIDIA/thrust/blob/main/.git-blame-ignore-revs) to this repo with the final SHA for this change so we can exclude these changes from blame, etc.

Another suggestion is to edit the commit author and change it to just `clang-format`. I like to clearly label these large automated changes as style-only in the metadata.

senior-zero:
> Before (or just after) merging this branch, add a file like [this one](https://github.com/NVIDIA/thrust/blob/main/.git-blame-ignore-revs) to this repo with the final SHA for this change so we can exclude these changes from blame, etc.
> 
> Another suggestion is to edit the commit author and change it to just `clang-format`. I like to clearly label these large automated changes as style-only in the metadata.

That's a great suggestion! Added .git-blame-ignore-revs file and changes the author.

PR #106, opened 2022-11-23
Fix bad implicit type conversions

aaronmondal:
Found because -Wsign-compare etc complained.

### DIFF
--- a/nvbench/detail/l2flush.cuh
+++ b/nvbench/detail/l2flush.cuh
@@ -31,11 +31,11 @@ struct l2flush
   {
     int dev_id{};
     NVBENCH_CUDA_CALL(cudaGetDevice(&dev_id));
-    NVBENCH_CUDA_CALL(cudaDeviceGetAttribute(&dev_id, cudaDevAttrL2CacheSize, dev_id));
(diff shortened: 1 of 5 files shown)


nvbench/detail/l2flush.cuh
```diff
@@ -31,7 +31,7 @@ struct l2flush
   {
     int dev_id{};
     NVBENCH_CUDA_CALL(cudaGetDevice(&dev_id));
-    NVBENCH_CUDA_CALL(cudaDeviceGetAttribute(&m_l2_size, cudaDevAttrL2CacheSize, dev_id));
+    NVBENCH_CUDA_CALL(cudaDeviceGetAttribute(&dev_id, cudaDevAttrL2CacheSize, dev_id));
```
karthikeyann, line 34: Should this be `&m_l2_size` itself?

aaronmondal: Oh yeah that is wrong. It should be `&m_l2_size`. But I believe I changed this because if `m_l2_size` is a `size_t`, this doesn't work. I'll find another way to write this.

aaronmondal: Ok new the updated variant should work :blush:

allisonvacanti:
Hi @aaronmondal, sorry it took way too long to get around to reviewing this. I'm going to close this since these issues were fixed in another commit since this PR was created, but I still appreciate the contribution!

PR #107, opened 2022-11-23
Fix wrongly defaulted move assignment operators

aaronmondal:
Found because -Wdefaulted-function-deleted complained.


allisonvacanti (commented):
These are intentional -- the classes are move-only, and should support move assignment.

Are these operators causing an issue?

senior-zero:
@aaronmondal I'm sorry, I've missed this PR and addressed the warnings as part of https://github.com/NVIDIA/nvbench/pull/139

PR #108, opened 2022-11-24
Fix filesystem detection for clang

aaronmondal:
Clang defines __GNUC__ but does not use experimental/filesystem.

### DIFF
--- a/nvbench/json_printer.cu
+++ b/nvbench/json_printer.cu
@@ -16,17 +16,23 @@
  *  limitations under the License.
  */
 
+#include <nvbench/json_printer.cuh>
+
 #include <nvbench/axes_metadata.cuh>
 #include <nvbench/benchmark_base.cuh>
 #include <nvbench/config.cuh>
-#include <nvbench/detail/throw.cuh>
 #include <nvbench/device_info.cuh>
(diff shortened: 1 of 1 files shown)


nvbench/json_printer.cu
```diff
@@ -16,24 +16,18 @@
  *  limitations under the License.
  */
 
-#include <nvbench/json_printer.cuh>
-
 #include <nvbench/axes_metadata.cuh>
 #include <nvbench/benchmark_base.cuh>
 #include <nvbench/config.cuh>
+#include <nvbench/detail/throw.cuh>
```
allisonvacanti, line 22: Let's avoid reordering the headers in this branch and just fix the clang bug

allisonvacanti (changes requested):
LGTM, but let's remove the unrelated changes. We'll be doing a large style sweep soon and I'd rather not thrash too much on these includes.

nvbench/json_printer.cu
aaronmondal: Reverted.

allisonvacanti (approved):

allisonvacanti:
Thanks!

PR #127, opened 2023-02-17
Propagate fmt requirement in conda envs

robertmaynard:
When we are inside a conda env the linker will be set to `ld.bfd` which will try to resolve all undefined symbols at time.

Since we could be using a shared library version of fmt we need it on the final link line of consumers


allisonvacanti (approved):
LGTM, tests pass.

PR #131, opened 2023-04-13
Add list-devices option

senior-zero:
This PR adds an option to get a list of devices in machine-readable format.


allisonvacanti (changes requested):
I like having `-l` print out both devices and benchmark info.

I usually run that to figure out the command line I need to run a specific benchmark on a specific device, and having it all convenient with `-l` is nice.

If we need to provide the benchmark and device details separately, I'd rather have less convenient options for those, since they'll most likely be used in scripts instead of interactively.

- `-l | --list`: list devices and benchmarks
- `--list-devices`: Just list device info
- `--list-benches`: Just list benchmark info

-------

Also, rather than having separate options for `list | jsonlist`, etc, this should be hooked into the existing format options, e.g. instead of `bench.exe --jsonlist`, it would be `bench.exe --json stdout --list` or `bench.exe -q --json foo.json --list-benches`.

allisonvacanti (approved):

PR #133, opened 2023-05-31
Handle use case where we are in a conda env but with a static fmt lib

robertmaynard:
Previously we presumed that a conda env meant we had a shared library version of fmt which isn't always the case. As a developer could be in a conda env that doesn't have the fmt package installed.


allisonvacanti (approved):
LGTM

PR #138, opened 2023-09-01
fix: add final mark in do_clone methods to avoid warnings

andmax:


miscco (approved):
Thanks a lot for the contribution 🎉

allisonvacanti (approved):

allisonvacanti:
Hi @andmax! Before we can merge this, we require all commits to be signed. Can you setup commit signing and remake this commit?

Info on signing is here: https://docs.github.com/articles/about-gpg/ (Using an SSH key instead of PGP is fine and generally easier to configure).

Updating the existing commit can be done with `git commit --amend -S` once the signing key is configured.

andmax:
Hi Allison, done!  I've gpg signed the commit amending it.
Thanks,
Andre.

allisonvacanti:
Great, thanks for the contribution!

PR #139, opened 2023-09-22
Fix NVBench Warnings

senior-zero:
NVBench is now compiled as part of the CCCL CI and produces a number of compilation warnings. This PR fixes those warnings.

### DIFF
--- a/nvbench/device_manager.cu
+++ b/nvbench/device_manager.cu
@@ -16,12 +16,13 @@
  *  limitations under the License.
  */
 
+#include <nvbench/device_manager.cuh>
+
 #include <cuda_runtime_api.h>
 
 #include <nvbench/cuda_call.cuh>
 #include <nvbench/detail/device_scope.cuh>
 #include <nvbench/detail/throw.cuh>
-#include <nvbench/device_manager.cuh>
 
 namespace nvbench
 {
--- a/nvbench/device_manager.cu
+++ b/nvbench/device_manager.cu
@@ -16,12 +16,12 @@
  *  limitations under the License.
  */
 
-#include <nvbench/device_manager.cuh>
+#include <cuda_runtime_api.h>
 
 #include <nvbench/cuda_call.cuh>
 #include <nvbench/detail/device_scope.cuh>
-
-#include <cuda_runtime_api.h>
+#include <nvbench/detail/throw.cuh>
+#include <nvbench/device_manager.cuh>
 
 namespace nvbench
 {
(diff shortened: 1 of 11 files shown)


nvbench/device_manager.cu
```diff
@@ -16,12 +16,12 @@
  *  limitations under the License.
  */
 
-#include <nvbench/device_manager.cuh>
+#include <cuda_runtime_api.h>
```
allisonvacanti, line 19: nit: NVBench puts the corresponding header as the first include, followed by a newline, for each translation unit. The other include changes here are fine, but let's leave the ordering as-is for consistently until we restyle the entire project.

nvbench/launch.cuh
```diff
@@ -43,9 +43,9 @@ struct launch
 
   // move-only
   launch(const launch &)            = delete;
-  launch(launch &&)                 = default;
+  launch(launch &&)                 = delete;
```
allisonvacanti, line 46: Ah, the const ref member is an issue...

The "move-only" comment needs to be updated.

nvbench/option_parser.cu
```diff
@@ -444,7 +444,7 @@ void option_parser::parse_range(option_parser::arg_iterator_t first,
       this->disable_blocking_kernel();
       first += 1;
     }
-    else if (arg == "--quiet" | arg == "-q")
+    else if (arg == "--quiet" || arg == "-q")
```
allisonvacanti, line 447: 🤦‍♀️

nvbench/printer_base.cuh
```diff
@@ -78,9 +78,9 @@ struct printer_base
 
   // move-only
   printer_base(const printer_base &)            = delete;
-  printer_base(printer_base &&)                 = default;
+  printer_base(printer_base &&)                 = delete;
```
allisonvacanti, line 81: Hrm, it would be nice to keep the printers movable.

Does CI still complain if these `=default` statements are moved to the `printer_base.cxx` file, similar to the `printer_base` destructor?

If we need to delete these, I won't object, but the comment will need to be updated.

allisonvacanti (approved):
Some minor issues / nitpicks / questions inline. This can be merged after these are addressed.

nvbench/device_manager.cu
senior-zero: Returned old headers order. Please, file an issue to teach clang-format the expected headers order.

nvbench/launch.cuh
senior-zero: Kept the move constructor and comment

nvbench/printer_base.cuh
senior-zero: Kept the move constructor and comment

PR #143, opened 2023-10-18
write_git_revision_file must be used in same CMakeLists as consumer

robertmaynard:
So we can't have this in the rapids-cmake init function.

### DIFF
--- a/nvbench/CMakeLists.txt
+++ b/nvbench/CMakeLists.txt
@@ -132,12 +132,20 @@ endif()
 # Use `nvbench_git_revision` but don't expose it as part of our export information
 # In CMake 3.27 this is easily done with $<COMPILE_ONLY>, before that we have to
 # manual copy the include path
+#
+# The call to `rapids_cmake_write_git_revision_file` must be in the same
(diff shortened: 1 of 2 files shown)


robertmaynard:
In local testing this resolves the build issues that CCCL is seeing with latest nvbench.

nvbench/CMakeLists.txt
```diff
@@ -132,12 +132,20 @@ endif()
...
 # In CMake 3.27 this is easily done with $<COMPILE_ONLY>, before that we have to
 # manual copy the include path
+#
+# The call to `rapids_cmake_write_git_revision_file` must be in the same
+# CMakeLists.txt as the consumer ( nvbench ) for CMake to get the dependency
+# graph correct.
+rapids_cmake_write_git_revision_file(
+    nvbench_git_revision
+    "${NVBench_BINARY_DIR}/nvbench/detail/git_revision.cuh"
+  )
```
miscco, line 142: Nitpick, some whitespace damage
```suggestion
  nvbench_git_revision
  "${NVBench_BINARY_DIR}/nvbench/detail/git_revision.cuh"
)
```

miscco (approved):

PR #148, opened 2023-11-14
Support static builds of nvbench with nvml enabled.

robertmaynard:
To do this we need to ensure that the nvml init handler is both contained in the library/executable that uses nvbench.

The original implementation fails since the singleton can be dropped since it has no usages. So instead we move to a function static which we ensure will always be used.

### DIFF
--- a/nvbench/internal/nvml.cuh
+++ b/nvbench/internal/nvml.cuh
@@ -33,7 +33,7 @@ namespace nvbench::nvml
 {
 
 // RAII struct that initializes and shuts down NVML
-// Beeds to be constructed and kept alive while using nvml
+// Needs to be constructed and kept alive while using nvml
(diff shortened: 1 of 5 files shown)


nvbench/internal/nvml.cuh
```diff
@@ -32,6 +32,16 @@
 namespace nvbench::nvml
 {
 
+// RAII struct that initializes and shuts down NVML
+// Beeds to be constructed and kept alive while using nvml
```
bdice, line 36:
```suggestion
// Needs to be constructed and kept alive while using nvml
```

bdice (approved):
This makes sense to me. Thanks!

allisonvacanti (approved):
lgtm

Issue #147, opened 2023-11-07
Consider `is_finished` customization


gevtushenko:
Currently, NVBench relies on a fixed [criteria](https://github.com/NVIDIA/nvbench/blob/57c4d42ba505d525a1a3815d2b2560450a4f83dd/nvbench/detail/measure_cold.cu#L102-L154) to understand when enough samples are collected. We can consider an API that would allow users to modify this criterion, accepting the set of collected samples and returning `true` if enough were collected. 

This API would make it:
1) easier to experiment with new criteria 
2) possible to offload decision to third party statistical packages

alliepiper:
Addressed by #151.

PR #151, opened 2024-01-05
Entropy-based stopping criterion

gevtushenko:
Closes https://github.com/NVIDIA/nvbench/issues/150 and https://github.com/NVIDIA/nvbench/issues/147. 

This PR adds new command line option `--stopping-criterion <criterion>` with two predefined criteria `stdrel` and `entropy` along with API for customization of the stopping criterion. The `nvbench/examples/custom_criterion.cu` illustrates how custom criteria can be added on per-run basis. This opens possibilities for performance CI improvements. One can now develop criteria that, for instance, collects large sample, store the sample size and then on each re-run of performance CI loads this number, leading to better stability. 

Apart from new API, `entropy` criterion is introduced. To enable it, it's sufficient to write `--stopping-criterion entropy`. The criterion computes cumulative entropy of the sample and stores it in an entropy window. Then, linear regression on the cumulative entropy window is computed. If the angle of the linear regression is small enough and coefficient of determination (R^2) is large enough, criterion believes that new samples will not introduce any new information and the sample is representative. Entropy criterion addresses concerns from https://github.com/NVIDIA/nvbench/issues/150 and https://github.com/NVIDIA/nvbench/issues/147 as well as significantly reduces variation of sample size, which is important for performance CI. Below is a plot of sample size distribution for `stdrel` and `entropy` criteria collected on `nvbench/examples/throughput.cu` that illustrates this point:

stdrel_vs_entropy

Below is an example where `stdrel` noticed small variance and decided to stop, but `entropy` noticed that entropy grows and kept sampling, discovering new modes:

large

large_4

Other times, `entropy` notices that new measurements do not introduce anything new to the sample and stops earlier:

small_2
small_3
small_5

Each criterion has its own set of parameters. Parameters like `--max-noise` and `--min-time` only affect `stdrel` criterion, whereas `--max-angle` and `--min-r2` are parameters of `entropy`. 

For now, `stdrel` stays as default criterion. Decision on switching the default criterion will be made after some field experience.

### DIFF
--- a/examples/custom_criterion.cu
+++ b/examples/custom_criterion.cu
@@ -1,0 +1,89 @@
+/*
+ *  Copyright 2023 NVIDIA Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 with the LLVM exception
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.
+ *
+ *  You may obtain a copy of the License at
+ *
+ *      http://llvm.org/foundation/relicensing/LICENSE.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include <nvbench/nvbench.cuh>
+
+// Grab some testing kernels from NVBench:
+#include <nvbench/test_kernels.cuh>
+
+// Thrust vectors simplify memory management:
+#include <thrust/device_vector.h>
+
+// Inherit from the stopping_criterion class:
+class fixed_criterion final : public nvbench::stopping_criterion 
+{
+  nvbench::int64_t m_max_samples{};
+  nvbench::int64_t m_num_samples{};
+
+public:
+  // Setup the criterion in the `initialize()` method:
+  virtual void initialize(const nvbench::criterion_params &params) override 
+  {
+    m_num_samples = 0;
+    m_max_samples = params.has_value("max-samples") ? params.get_int64("max-samples") : 42;
+  }
+
+  // Process new measurements in the `add_measurement()` method:
+  virtual void add_measurement(nvbench::float64_t /* measurement */) override
+  {
+    m_num_samples++;
+  }
+
+  // Check if the stopping criterion is met in the `is_finished()` method:
+  virtual bool is_finished() override
+  {
+    return m_num_samples >= m_max_samples;
+  }
+
+  // Describe criterion parameters in the `get_params()` method:
+  virtual const params_description &get_params() const override
+  {
+    static const params_description desc{
+      {"max-samples", nvbench::named_values::type::int64}
+    };
+    return desc;
+  }
+};
+
+// Register the criterion with NVBench:
+static bool registered = //
+  nvbench::criterion_registry::register_criterion("fixed",
+                                                  std::make_unique<fixed_criterion>());
+
+void throughput_bench(nvbench::state &state)
+{
+  // Allocate input data:
+  const std::size_t num_values = 64 * 1024 * 1024 / sizeof(nvbench::int32_t);
+  thrust::device_vector<nvbench::int32_t> input(num_values);
+  thrust::device_vector<nvbench::int32_t> output(num_values);
+
+  // Provide throughput information:
+  state.add_element_count(num_values, "NumElements");
+  state.add_global_memory_reads<nvbench::int32_t>(num_values, "DataSize");
+  state.add_global_memory_writes<nvbench::int32_t>(num_values);
+
+  state.set_stopping_criterion("fixed");
+
+  state.exec(nvbench::exec_tag::no_batch, [&input, &output, num_values](nvbench::launch &launch) {
+    nvbench::copy_kernel<<<256, 256, 0, launch.get_stream()>>>(
+      thrust::raw_pointer_cast(input.data()),
+      thrust::raw_pointer_cast(output.data()),
+      num_values);
+  });
+}
+NVBENCH_BENCH(throughput_bench);
--- a/examples/custom_criterion.cu
+++ b/examples/custom_criterion.cu
@@ -24,14 +24,14 @@
 // Thrust vectors simplify memory management:
 #include <thrust/device_vector.h>
 
-// Inherit from the stopping_criterion class:
-class fixed_criterion final : public nvbench::stopping_criterion 
+// Inherit from the stopping_criterion_base class:
+class fixed_criterion final : public nvbench::stopping_criterion_base
 {
   nvbench::int64_t m_num_samples{};
 
 public:
   fixed_criterion()
-      : nvbench::stopping_criterion{"fixed", {{"max-samples", nvbench::int64_t{42}}}}
+      : nvbench::stopping_criterion_base{"fixed", {{"max-samples", nvbench::int64_t{42}}}}
   {}
 
   // Process new measurements in the `add_measurement()` method:
--- a/examples/custom_criterion.cu
+++ b/examples/custom_criterion.cu
@@ -27,15 +27,19 @@
 // Inherit from the stopping_criterion class:
 class fixed_criterion final : public nvbench::stopping_criterion 
 {
+  nvbench::criterion_params m_params{};
   nvbench::int64_t m_max_samples{};
   nvbench::int64_t m_num_samples{};
 
 public:
+  fixed_criterion() : nvbench::stopping_criterion{"fixed"} {}
+
   // Setup the criterion in the `initialize()` method:
   virtual void initialize(const nvbench::criterion_params &params) override 
   {
+    m_params = params;
     m_num_samples = 0;
-    m_max_samples = params.has_value("max-samples") ? params.get_int64("max-samples") : 42;
+    m_max_samples = m_params.has_value("max-samples") ? m_params.get_int64("max-samples") : 42;
   }
 
   // Process new measurements in the `add_measurement()` method:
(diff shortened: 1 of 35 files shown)


nvbench/detail/entropy_criterion.cxx
```diff
@@ -0,0 +1,153 @@
...
+
+bool entropy_criterion::is_finished()
+{
+  if (m_entropy_tracker.size() < 2)
+  {
+    return false;
+  }
+
+  // About 20% faster than the stdrel method
+  if (m_total_samples % 2 != 0)
```
allisonvacanti, line 120: Is an even number of samples a requirement for the entropy calc? Or is this just to reduce the overhead of compute the entropy?

Can you add a comment elaborating on this?

nvbench/stopping_criterion.cuh
```diff
@@ -0,0 +1,86 @@
...
+  [[nodiscard]] nvbench::int64_t get_int64(const std::string &name) const;
+  [[nodiscard]] nvbench::float64_t get_float64(const std::string &name) const;
+};
+
+/**
+ * Stopping criterion interface
+ */
+class stopping_criterion
+{
+public:
```
allisonvacanti, line 60: Consider adding `name` and `params` members to better encapsulate the stopping criterion as a single, self-contained unit.

nvbench/stopping_criterion.cuh
```diff
@@ -0,0 +1,86 @@
...
+   * Check if the criterion has been met for all measurements processed by `add_measurement`
+   */
+  virtual bool is_finished() = 0;
+
+  using params_description = std::vector<std::pair<std::string, nvbench::named_values::type>>;
+
+  /**
+   * Return the parameter names and types for this criterion
+   */
+  virtual const params_description &get_params() const = 0;
```
allisonvacanti, line 83: Rename to `get_params_description`.

allisonvacanti (changes requested):

examples/custom_criterion.cu
```diff
@@ -0,0 +1,89 @@
...
+  {
+    static const params_description desc{
+      {"max-samples", nvbench::named_values::type::int64}
+    };
+    return desc;
+  }
+};
+
+// Register the criterion with NVBench:
+static bool registered = //
```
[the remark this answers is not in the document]
allisonvacanti: If the name of the criterion is made a member of the criterion class, the string can be omitted.

nvbench/benchmark_base.cuh
```diff
@@ -181,22 +182,24 @@ struct benchmark_base
   }
   /// @}
 
-  /// Accumulate at least this many seconds of timing data per measurement. @{
-  [[nodiscard]] nvbench::float64_t get_min_time() const { return m_min_time; }
+  /// Accumulate at least this many seconds of timing data per measurement.
+  /// Only applies to `stdrel` stopping criterion. @{
+  [[nodiscard]] nvbench::float64_t get_min_time() const { return m_criterion_params.get_float64("min-time"); }
```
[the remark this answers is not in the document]
gevtushenko: Please, create an issue and deprecate it as a separate PR

nvbench/criterion_registry.cuh
```diff
@@ -0,0 +1,49 @@
...
+
+public:
+  static criterion_registry &instance();
+
+  static nvbench::stopping_criterion* get(const std::string& name);
+
+  static bool register_criterion(std::string name,
+                                 std::unique_ptr<nvbench::stopping_criterion> criterion);
+
+  static nvbench::stopping_criterion::params_description get_params_description();
```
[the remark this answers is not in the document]
gevtushenko: Please, file an issue to investigate this idea later.

nvbench/detail/entropy_criterion.cxx
```diff
@@ -0,0 +1,156 @@
...
+    auto key = measurement;
+    constexpr bool bin_keys = false;
+
+    if (bin_keys) 
+    {
+      const auto resolution_us = 0.5;
+      const auto resulution_s = resolution_us / 1'000'000;
+      const auto epsilon = resulution_s * 2;
+      key = std::round(key / epsilon) * epsilon;
+    }
```
julie-is-late, line 93: Definitely good to have a binning option available to investigate. 
 
I think that the image example from the MR description for `cub.bench.scan.exclusive.sum::base` is a good example of something that I think might benefit from it. I'm not the most familiar with the specifics of cuda timing interface, however having worked with data from the perfmon hardware I am assuming it is based on, I can say that there is definitely a lower limit to the data granularity simply due to the polling rate on the hardware. But beyond even the practicality of measuring things this precise, I think it's worth asking if that data is useful to the user? Just something to think about.

nvbench/detail/entropy_criterion.cuh
```diff
@@ -0,0 +1,54 @@
...
+{
+  // parameters
+  nvbench::float64_t m_max_angle{0.048};
+  nvbench::float64_t m_min_r2{0.36};
+
+  // state
+  nvbench::int64_t m_total_samples{};
+  nvbench::float64_t m_total_cuda_time{};
+  std::vector<std::pair<nvbench::float64_t, nvbench::int64_t>> m_freq_tracker;
+  nvbench::detail::ring_buffer<nvbench::float64_t> m_entropy_tracker{299};
```
julie-is-late, line 40: I wonder if you don't want to make the default size for `m_entropy_tracker` user configurable for testing like you do for max_angle even if it means having to allocate a new buffer. How sensitive is the program to the size of this lookback?

gevtushenko: > How sensitive is the program to the size of this lookback?

The criterion is sensitive to the window size. As of now, it's a bit complicated to make this user configurable, but I agree that it should be done at some point. Added a TODO.

nvbench/detail/entropy_criterion.cxx
```diff
@@ -0,0 +1,156 @@
...
+bool entropy_criterion::is_finished()
+{
+  if (m_entropy_tracker.size() < 2)
+  {
+    return false;
+  }
+
+  // Even number of samples is used to reduce the overhead and not required to compute entropy.
+  // This makes `is_finished()` about 20% faster than corresponding stdrel method.
+  if (m_total_samples % 2 != 0)
```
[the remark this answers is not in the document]
gevtushenko: I haven't considered this idea, thank you! Filed a separate issue to investigate it https://github.com/NVIDIA/nvbench/issues/154.

nvbench/detail/statistics.cuh
```diff
@@ -56,8 +58,124 @@ ValueType standard_deviation(Iter first, Iter last, ValueType mean)
...
-                        static_cast<ValueType>((num - 1));
+                        static_cast<ValueType>((num - 1)); // Bessel’s correction
   return std::sqrt(variance);
 }
 
+template <class It>
+nvbench::float64_t compute_mean(It first, It last)
+{
+  const auto n = std::distance(first, last);
+  return std::accumulate(first, last, 0.0) / static_cast<nvbench::float64_t>(n);
```
julie-is-late, line 69: This has a divide by 0 error if the iterator has a size 0. Previously this was guarded by the size check at line 84, but now this division is done before that is checked.

examples/custom_criterion.cu
```diff
@@ -0,0 +1,89 @@
...
+
+public:
+  fixed_criterion() : nvbench::stopping_criterion{"fixed"} {}
+
+  // Setup the criterion in the `initialize()` method:
+  virtual void initialize(const nvbench::criterion_params &params) override 
+  {
+    m_params = params;
+    m_num_samples = 0;
+    m_max_samples = m_params.has_value("max-samples") ? m_params.get_int64("max-samples") : 42;
```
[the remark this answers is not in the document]
gevtushenko: I reworked the API a bit, now there's no `get_params_description` and the content of the parameters is the source of truth.

nvbench/stopping_criterion.cuh
```diff
@@ -0,0 +1,120 @@
...
+  [[nodiscard]] bool has_value(const std::string &name) const;
+  [[nodiscard]] nvbench::int64_t get_int64(const std::string &name) const;
+  [[nodiscard]] nvbench::float64_t get_float64(const std::string &name) const;
+  [[nodiscard]] std::string get_string(const std::string &name) const;
+};
+
+/**
+ * Stopping criterion interface
+ */
+class stopping_criterion
```
allisonvacanti, line 73: The other pure virtual base classes in NVBench all end in `_base`. Let's rename this to `stopping_criterion_base` for consistency.

nvbench/option_parser.cu
```diff
@@ -504,16 +514,29 @@ void option_parser::parse_range(option_parser::arg_iterator_t first,
...
     {
       check_params(1);
       this->update_float64_prop(first[0], first[1]);
       first += 2;
     }
     else
-    {
-      NVBENCH_THROW(std::runtime_error, "Unrecognized command-line argument: `{}`.", arg);
+    { // Try criterion params
+      std::string_view name(first[0].c_str() + 2, first[0].size() - 2);
```
allisonvacanti, line 525: Check if the length of the arg > 3 and that the first two chars are both `-` before manipulating the pointer like this.

nvbench/stopping_criterion.cuh
```diff
@@ -0,0 +1,120 @@
...
+  criterion_params m_params;
+
+public:
+  /**
+   * @param name Unique name of the criterion
+   * @param params Default values for all parameters of the criterion
+   */
+  explicit stopping_criterion(std::string name, criterion_params params)
+      : m_name{std::move(name)}
+      , m_params{params}
```
allisonvacanti, line 86:
```suggestion
      , m_params{std::move(params)}
```

allisonvacanti (approved):
LGTM -- Thanks for this, it should be a huge improvement!

PR #157, opened 2024-04-03
Set `CUDA_MODULE_LOADING=EAGER` before `main`.

### LINKED ISSUE #136: Async benchmarks always deadlock
The recent switch to lazy loading by default in CTK 12.2 seems to have broken the async benchmarks. This can be reproduced by `nvbench.example.axes`. The deadlock can be fixed by `CUDA_MODULE_LOADING=EAGER`. We should incorporate this information into the error message or set the variable ourselves.

alliepiper:
Fixes #136


nvbench/main.cuh
```diff
@@ -25,8 +25,16 @@
 #include <nvbench/option_parser.cuh>
 #include <nvbench/printer_base.cuh>
 
+#include <cstdlib>
 #include <iostream>
 
+static int _nvbench_env [[maybe_unused]] = []() -> int
+{
+  // See NVIDIA/NVBench#136
+  int retval = setenv("CUDA_MODULE_LOADING", "EAGER", 1);
```
gevtushenko, line 34: important: `setenv` is a POSIX extension. This change will break MSVC.

nvbench/main.cuh
```diff
@@ -25,8 +25,16 @@
 #include <nvbench/option_parser.cuh>
 #include <nvbench/printer_base.cuh>
 
+#include <cstdlib>
 #include <iostream>
 
+static int _nvbench_env [[maybe_unused]] = []() -> int
```
gevtushenko, line 31: question: Initialization of global variables has undefined order. Since we control the main function, having the `setenv` right at the beginning of main (and before any of the CUDA calls) seems like more sound approach. Is there any reason to set the environment variable before main?

gevtushenko (changes requested):

nvbench/main.cuh
alliepiper: > Is there any reason to set the environment variable before main?

I tried that first, but setting the env var at the start of main does not work -- it appears that CUDA is initialized prior to main. Using this static initialization works in my tests, but I am also concerned about initialization order and want to test it on a larger set of benchmarks to see if this is workable solution. Otherwise we may need to find a way to restart the CUDA context or something.

nvbench/main.cuh
alliepiper: Good catch!

nvbench/main.cuh
alliepiper: Updated to remove the pre-main usage of CUDA (thanks @pciolkosz!) and moved the setenv into main -- which works now :)

nvbench/option_parser.cu
```diff
@@ -781,6 +782,8 @@ catch (std::exception &e)
 
 void option_parser::replay_global_args()
 {
+  // Initialize benchmark with all devices:
+  m_benchmarks.back()->set_devices(nvbench::device_manager::get().get_devices());
```
gevtushenko, line 786: important: it seems the contract is that devices are always set before benchmark is used. It took me a while to track setting and getting devices in current codebase to see if this change is safe.  From the language perspective, nothing guarantees that devices are set before they are used. I think this makes code fragile. I'm not sure how the right refactoring would look like, but we could at least use optional:
```
std::optional<std::vector<nvbench::device_info>> m_devices;
```
This way we'll get an exception if devices are referenced before they are set.

gevtushenko (approved):

nvbench/option_parser.cu
alliepiper: I agree that this is fragile, but I'd rather not have an empty device list be an error condition. If a user programmatically filters devices out and ends up with none left in a particular benchmark, skipping the benchmark instead of throwing an error seems least surprising IMO.

I think I see a clean way of doing this without being too invasive. Since the benchmarks are always cloned from the `benchmark_manager`, we can setup benchmark defaults in a `benchmark_manager::initialize` function from `main`. Then the devices will get cloned with the full set of default devices, before the `option_parser` potentially reduces the coverage.

PR #162, opened 2024-04-06
Cleanup_main

alliepiper:
Moves as much of the guts of `main` into actual functions as possible.

The macros are now easier to reuse and redefine for users who want to customize main.
Existing macro usage should continue to work.

### DIFF
--- a/nvbench/main.cuh
+++ b/nvbench/main.cuh
@@ -28,14 +28,50 @@
 #include <cstdlib>
 #include <iostream>
 
+// Advanced users can rebuild NVBench's `main` function using the macros in this file, or replace
+// them with customized implementations. The default implementation is provided below.
+
+#ifndef NVBENCH_MAIN
 #define NVBENCH_MAIN                                                                               \
-  int main(int argc, char const *const *argv)                                                      \
-  try                                                                                              \
-  {                                                                                                \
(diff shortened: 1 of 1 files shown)


alliepiper:
Need to add tests for the customization points I see people using on github

alliepiper:
Might pursue later, but this would break existing usecases, eg cudf currently does:

```

#include <nvbench/main.cuh>

#undef NVBENCH_MAIN_PARSE
#define NVBENCH_MAIN_PARSE(argc, argv)                                                             \
  nvbench::option_parser parser;                                                                   \
  std::vector<std::string> m_args;                                                                 \
  for (int i = 0; i < argc; ++i)                                                                   \
  {                                                                                                \
    std::string arg = argv[i];                                                                     \
    if (arg == "foo")                                                                              \
    {                                                                                              \
      i += 2;                                                                                      \
    }                                                                                              \
    else if (arg == "bar")                                                                         \
    {                                                                                              \
      i += 2;                                                                                      \
    }                                                                                              \
    else                                                                                           \
    {                                                                                              \
      m_args.push_back(arg);                                                                       \
    }                                                                                              \
  }                                                                                                \
  parser.parse(m_args)

NVBENCH_MAIN
```

Since `NVBENCH_MAIN` now just calls a function, the custom `PARSE(...)` macro is no longer used.

PR #173, opened 2024-05-29
Added cudaGetLastError() calls to reset benchmarking kernel errors (issue 88).

psvvsp:
Created the NVBENCH_CUDA_CALL_RESET_ERROR() macro and use it to wrap cudaStreamSynchronize() calls.

### DIFF
--- a/nvbench/cuda_call.cuh
+++ b/nvbench/cuda_call.cuh
@@ -24,12 +24,14 @@
 #include <string>
 
 /// Throws a std::runtime_error if `call` doesn't return `cudaSuccess`.
+/// Resets the error with cudaGetLastError().
 #define NVBENCH_CUDA_CALL(call)                                                                    \
   do                                                                                               \
   {                                                                                                \
     const cudaError_t nvbench_cuda_call_error = call;                                              \
     if (nvbench_cuda_call_error != cudaSuccess)                                                    \
     {                                                                                              \
+      cudaGetLastError();                                                                          \
       nvbench::cuda_call::throw_error(__FILE__, __LINE__, #call, nvbench_cuda_call_error);         \
     }                                                                                              \
   } while (false)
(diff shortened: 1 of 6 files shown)


nvbench/cuda_call.cuh
```diff
@@ -57,6 +57,21 @@
     }                                                                                              \
   } while (false)
 
+/// Throws a std::runtime_error if `call` doesn't return `cudaSuccess`.
+/// Resets the error with cudaGetLastError().
+#define NVBENCH_CUDA_CALL_RESET_ERROR(call)                                                        \
```
alliepiper, line 62: Rather than add a new macro, let's add the `cudaGetLastError` call to the existing macros, since we'll always want to clear them.

testing/reset_error.cu
```diff
@@ -0,0 +1,29 @@
...
+}
+
+int main()
+{ 
+  multiply5<<<256, 256>>>(nullptr, nullptr);
+
+  try
+  {
+    NVBENCH_CUDA_CALL_RESET_ERROR(cudaStreamSynchronize(0));
+    return 1;
```
alliepiper, line 21: Add an assert here that always fails if the the NVBENCH_CUDA_CALL macro doesn't throw.

testing/reset_error.cu
```diff
@@ -0,0 +1,29 @@
...
+  }
+  catch (const std::runtime_error &)
+  {
+    ASSERT(cudaGetLastError() == cudaError_t::cudaSuccess);
+  }
+
+  return 0;
+}
```
alliepiper, line 29: Add missing newline.

alliepiper (changes requested):
Thanks for the PR! I've pointed out some changes inline, but this can go in once those are addressed.

nvbench/cuda_call.cuh
psvvsp: Deleted NVBENCH_CUDA_CALL_RESET_ERROR and moved cudaGetLastError() call to NVBENCH_CUDA_CALL macro.
Other macros don't need cudaGetLastError() because
1) Driver API calls don't set runtime last error
2) NVBENCH_CUDA_CALL_NOEXCEPT immediately exits the program.

testing/reset_error.cu
psvvsp: Done.

testing/reset_error.cu
psvvsp: Done.

alliepiper (approved):

alliepiper:
LGTM, thanks for the PR!

PR #205, opened 2025-04-08
typo in benchmarks.md

gonidelis:


docs/benchmarks.md
```diff
@@ -435,9 +435,9 @@ NVBENCH_BENCH(my_cpu_benchmark)
   .set_is_cpu_only(true); // Mark as CPU-only.
 ```
 
-The optional `nvbench::exec_tag::no_gpu` hint may be used to reduce tbe compilation time and
+The optional `nvbench::exec_tag::no_gpu` hint may be used to reduce the compilation time and
 binary size of CPU-only benchmarks. An error is emitted at runtime if this tag is used while
-`is_cpu_only` is false.
+`set_is_cpu_only` is false.
```
gonidelis, line 440: ah, `is_cpu_only` is the name of the property. feel free to scrap these changes.

docs/benchmarks.md
```diff
@@ -435,9 +435,9 @@ NVBENCH_BENCH(my_cpu_benchmark)
   .set_is_cpu_only(true); // Mark as CPU-only.
 ```
 
-The optional `nvbench::exec_tag::no_gpu` hint may be used to reduce tbe compilation time and
```
gonidelis, line 438: @alliepiper  but there is still this :/

gonidelis: I reset hard the branch to the commit that only fixes this but because PR is closed it doesn't show.

Issue #193, opened 2024-11-25
Check average clock frequency during benchmarks


bernhardmgruber:
Sometimes, benchmark systems are unstable due to external factors and GPUs cannot keep up their clock frequency during a benchmark. This leads to wrong results.

NVBench should monitor the clock frequency during benchmarking and detect such conditions. One way is to query the global timer and SM block before and after the benchmark, and compute the average frequency:
```c++
__global__ void get_timestamps_kernel(uint64_t* global_timestamp, uint64_t* sm0_timestamp) {
  uint32_t smid;
  asm volatile("mov.u32 %0, %%smid;" : "=r"(smid));
  if (smid == 0) {
    uint64_t gts, lts;
    asm volatile("mov.u64 %0, %%globaltimer;" : "=l"(gts));
    lts = clock64();

    *global_timestamp = gts;
    *sm0_timestamp    = lts;
  }
}

ulong2 get_timestamps() {
  const int device_id = 0;
  int num_sms         = 0;
  CUDA_CHECK(cudaDeviceGetAttribute(&num_sms, cudaDevAttrMultiProcessorCount, device_id));

  uint64_t* timestamps;
  CUDA_CHECK(cudaHostAlloc(&timestamps, 2 * sizeof(uint64_t), cudaHostAllocMapped));
  get_timestamps_kernel<<<num_sms, 1>>>(&timestamps[0], &timestamps[1]);
  CUDA_CHECK(cudaDeviceSynchronize());
  ulong2 ret{timestamps[0], timestamps[1]};
  CUDA_CHECK(cudaFreeHost(timestamps));
  return ret;
}

float measure_clock_frequency(std::function<void(void)> f) {
  ulong2 ts0 = get_timestamps();
  f();
  ulong2 ts1              = get_timestamps();
  uint64_t elapsed_ns     = ts1.x - ts0.x;
  uint64_t elapsed_clocks = ts1.y - ts0.y;
  float clock_rate        = float(elapsed_clocks) / float(elapsed_ns) * 1000000.f;
  return clock_rate;
}
```
where `f` launches the kernel to benchmark. If the computed `clock_rate` is off from the expected value, we should issue a warning.

fbusato:
this can be also achieved with the `nvml` library

[...]

PR #206, opened 2025-04-09
Discard measurements while GPU is throttling

gevtushenko:
Fixes https://github.com/NVIDIA/nvbench/issues/193

This PR adds a few options:
1. `--throttle-threshold 0.8` (80%) - sets percentage of peak frequency that's considered throttling. If SM frequency drops below this value, warning is issued. 
2. `--throttle-recovery-delay 5.0` - waits for 5 seconds to let GPU cool down if throttling took place. 
3. `--discard-on-throttle` - discards measurements taken while GPU was throttling. 

These end up having positive effect on the benchmarks. For `select.if` benchmark, upstream NVBench yields:

| T{ct} | OffsetT{ct} | InPlace{ct} |   Elements{io}   | Entropy | Samples | CPU Time | Noise | GPU Time | Noise |  Elem/s  | GlobalMem BW | BWUtil |
|-------|-------------|-------------|------------------|---------|---------|----------|-------|----------|-------|----------|--------------|--------|
|    I8 |         I64 |        true | 2^28 = 268435456 |   1.000 |  10000x | 1.068 ms | 5.34% | 1.056 ms | 3.22% | 254.310G | 507.628 GB/s | 52.87% |

With `--throttle-threshold 0.8 --throttle-recovery-delay 5.0 --discard-on-throttle` resulting noise twice as low and the resulting kernel perf is much higher:

| T{ct} | OffsetT{ct} | InPlace{ct} |   Elements{io}   | Entropy | Samples |  CPU Time  | Noise |  GPU Time  | Noise |  Elem/s  | GlobalMem BW | BWUtil |
|-------|-------------|-------------|------------------|---------|---------|------------|-------|------------|-------|----------|--------------|--------|
|    I8 |         I64 |        true | 2^28 = 268435456 |   1.000 |  10000x | 985.749 us | 4.85% | 972.956 us | 1.54% | 275.897G | 550.716 GB/s | 57.36% |

Further increasing the throttle threshold to 90% yields even lower noise and higher perf:

| T{ct} | OffsetT{ct} | InPlace{ct} |   Elements{io}   | Entropy | Samples |  CPU Time  | Noise |  GPU Time  | Noise |  Elem/s  | GlobalMem BW | BWUtil |
|-------|-------------|-------------|------------------|---------|---------|------------|-------|------------|-------|----------|--------------|--------|
|    I8 |         I64 |        true | 2^28 = 268435456 |   1.000 |  10000x | 965.393 us | 4.31% | 953.132 us | 1.02% | 281.635G | 562.171 GB/s | 58.55% |

The effect can be seen on underlying distributions, we get faster kernels that are not affected by throttling caused by NVBench invoking the kernel thousand times:

distributions

Comparison of frequencies and temperatures with / without throttling control:

frequency

temperature

### DIFF
--- a/docs/cli_help.md
+++ b/docs/cli_help.md
@@ -133,11 +133,6 @@
   * Applies to the most recent `--benchmark`, or all benchmarks if specified
     before any `--benchmark` arguments.
 
-* `--discard-on-throttle`
-  * Discard measurements if the GPU is throttled.
-  * Applies to the most recent `--benchmark`, or all benchmarks if specified
-    before any `--benchmark` arguments.
-
 * `--throttle-threshold <value>`
   * Set the GPU throttle threshold as percentage of the peak clock rate.
   * Default is 75%.
--- a/docs/cli_help.md
+++ b/docs/cli_help.md
@@ -140,13 +140,13 @@
 
 * `--throttle-threshold <value>`
   * Set the GPU throttle threshold as percentage of the peak clock rate.
-  * Default is 0.75 (75%).
+  * Default is 75%.
   * Applies to the most recent `--benchmark`, or all benchmarks if specified
     before any `--benchmark` arguments.
 
 * `--throttle-recovery-delay <value>`
   * Set the GPU throttle recovery delay in seconds.
-  * Default is 0.5 seconds.
+  * Default is 0.05 seconds.
   * Applies to the most recent `--benchmark`, or all benchmarks if specified
     before any `--benchmark` arguments.
 
--- a/docs/cli_help.md
+++ b/docs/cli_help.md
@@ -133,6 +133,23 @@
   * Applies to the most recent `--benchmark`, or all benchmarks if specified
     before any `--benchmark` arguments.
 
+* `--discard-on-throttle`
+  * Discard measurements if the GPU is throttled.
+  * Applies to the most recent `--benchmark`, or all benchmarks if specified
+    before any `--benchmark` arguments.
+
+* `--throttle-threshold <value>`
+  * Set the GPU throttle threshold as percentage of the peak clock rate.
+  * Default is 0.75 (75%).
+  * Applies to the most recent `--benchmark`, or all benchmarks if specified
+    before any `--benchmark` arguments.
+
+* `--throttle-recovery-delay <value>`
+  * Set the GPU throttle recovery delay in seconds.
+  * Default is 0.5 seconds.
+  * Applies to the most recent `--benchmark`, or all benchmarks if specified
+    before any `--benchmark` arguments.
+
 * `--run-once`
   * Only run the benchmark once, skipping any warmup runs and batched
     measurements.
--- a/nvbench/benchmark_base.cuh
+++ b/nvbench/benchmark_base.cuh
@@ -309,8 +309,8 @@ protected:
   nvbench::float64_t m_skip_time{-1.};
   nvbench::float64_t m_timeout{15.};
 
-  nvbench::float32_t m_throttle_threshold{0.75f};     // [% of peak SM clock rate]
-  nvbench::float32_t m_throttle_recovery_delay{0.0f}; // [seconds]
+  nvbench::float32_t m_throttle_threshold{0.75f};      // [% of peak SM clock rate]
+  nvbench::float32_t m_throttle_recovery_delay{0.05f}; // [seconds]
   bool m_discard_on_throttle{false};
 
   nvbench::criterion_params m_criterion_params;
--- a/nvbench/benchmark_base.cuh
+++ b/nvbench/benchmark_base.cuh
@@ -247,12 +247,39 @@ struct benchmark_base
   }
   /// @}
 
-  [[nodiscard]] nvbench::criterion_params& get_criterion_params() { return m_criterion_params; }
-  [[nodiscard]] const nvbench::criterion_params& get_criterion_params() const { return m_criterion_params; }
+  [[nodiscard]] nvbench::float32_t get_throttle_threshold() const { return m_throttle_threshold; }
+
+  void set_throttle_threshold(nvbench::float32_t throttle_threshold)
+  {
+    m_throttle_threshold = throttle_threshold;
+  }
+
+  [[nodiscard]] nvbench::float32_t get_throttle_recovery_delay() const
+  {
+    return m_throttle_recovery_delay;
+  }
+
+  void set_throttle_recovery_delay(nvbench::float32_t throttle_recovery_delay)
+  {
+    m_throttle_recovery_delay = throttle_recovery_delay;
+  }
+
+  [[nodiscard]] bool get_discard_on_throttle() const { return m_discard_on_throttle; }
+
+  void set_discard_on_throttle(bool discard_on_throttle)
+  {
+    m_discard_on_throttle = discard_on_throttle;
+  }
+
+  [[nodiscard]] nvbench::criterion_params &get_criterion_params() { return m_criterion_params; }
+  [[nodiscard]] const nvbench::criterion_params &get_criterion_params() const
+  {
+    return m_criterion_params;
+  }
 
   /// Control the stopping criterion for the measurement loop.
   /// @{
-  [[nodiscard]] const std::string& get_stopping_criterion() const { return m_stopping_criterion; }
+  [[nodiscard]] const std::string &get_stopping_criterion() const { return m_stopping_criterion; }
   benchmark_base &set_stopping_criterion(std::string criterion)
   {
     m_stopping_criterion = std::move(criterion);
@@ -281,6 +308,10 @@ protected:
 
   nvbench::float64_t m_skip_time{-1.};
   nvbench::float64_t m_timeout{15.};
+
+  nvbench::float32_t m_throttle_threshold{0.75f};     // [% of peak SM clock rate]
+  nvbench::float32_t m_throttle_recovery_delay{0.0f}; // [seconds]
+  bool m_discard_on_throttle{false};
 
   nvbench::criterion_params m_criterion_params;
   std::string m_stopping_criterion{"stdrel"};
--- a/nvbench/benchmark_base.cuh
+++ b/nvbench/benchmark_base.cuh
@@ -264,13 +264,6 @@ struct benchmark_base
     m_throttle_recovery_delay = throttle_recovery_delay;
   }
 
-  [[nodiscard]] bool get_discard_on_throttle() const { return m_discard_on_throttle; }
-
-  void set_discard_on_throttle(bool discard_on_throttle)
-  {
-    m_discard_on_throttle = discard_on_throttle;
-  }
-
   [[nodiscard]] nvbench::criterion_params &get_criterion_params() { return m_criterion_params; }
   [[nodiscard]] const nvbench::criterion_params &get_criterion_params() const
   {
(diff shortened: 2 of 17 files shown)


ahendriksen (commented):
Really cool! Thanks for adding this Georgii!

Would it also be possible to report the average frequency in the output table?

Something like this (note second-to-last column):


| Operation | Compression  |   Initialization    | grid_dim | block_dim | occupancy | # registers | smem kb | register_usage | Samples | CPU Time | Noise | GPU Time | Noise |  Elem/s  | GlobalMem BW | BWUtil  | Eff. Clock Freq | bytes_in_flight_SM |
|-----------|--------------|---------------------|----------|-----------|-----------|-------------|---------|----------------|---------|----------|-------|----------|-------|----------|--------------|---------|-----------------|--------------------|
|  op_write | uncompressed |           init_zero |   262144 |       256 |         8 |          32 |       0 |        100.00% |    383x | 1.311 ms | 0.42% | 1.306 ms | 0.22% | 411.149G |   3.288 TB/s |  98.07% |          1.979G |            0.000 B |
|  op_write | uncompressed |       init_constant |   262144 |       256 |         8 |          32 |       0 |        100.00% |    383x | 1.311 ms | 0.42% | 1.307 ms | 0.23% | 411.086G |   3.287 TB/s |  98.05% |          1.979G |            0.000 B |

docs/cli_help.md
```diff
@@ -133,6 +133,23 @@
...
     before any `--benchmark` arguments.
 
+* `--discard-on-throttle`
+  * Discard measurements if the GPU is throttled.
+  * Applies to the most recent `--benchmark`, or all benchmarks if specified
+    before any `--benchmark` arguments.
+
+* `throttle-threshold <value>`
+  * Set the GPU throttle threshold as percentage of the peak clock rate.
+  * Default is 0.5 (50%).
```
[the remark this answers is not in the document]
gevtushenko: That's a typo, I've set the default to 80% in actual code. Will update to 75%, thanks! Regarding separate column on the output, I think it makes sense, will add it.

nvbench/timestamps_kernel.cuh
```diff
@@ -0,0 +1,45 @@
...
+#pragma once
+
+#include <nvbench/types.cuh>
+
+namespace nvbench
+{
+
+struct cuda_stream;
+
+struct timestamps_kernel
```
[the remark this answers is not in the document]
alliepiper: suggestion: also consider moving this into `detail/`/`nvbench::detail`, since I wouldn't expect users to need this.

Admittedly I've been sloppy with this in the past and need to do a big cleanup to separate things into `detail/`, so feel free to punt on this for now.

nvbench/gpu_frequency.cuh
```diff
@@ -0,0 +1,53 @@
...
+
+#include <nvbench/timestamps_kernel.cuh>
+#include <nvbench/types.cuh>
+
+namespace nvbench
+{
+
+struct cuda_stream;
+
+struct gpu_frequency
```
alliepiper, line 29: minor: Docstrings/comments on this class and its usage would be helpful long-term.

suggestion: Consider moving to `detail` folder/namespace (see comment on the timestamp class, too, no worries if you want to leave it here for now).

nvbench/timestamps_kernel.cu
```diff
@@ -0,0 +1,83 @@
...
+ *      http://llvm.org/foundation/relicensing/LICENSE.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include <cuda/std/chrono>
```
alliepiper, line 19: minor: Unused.

nvbench/timestamps_kernel.cu
```diff
@@ -0,0 +1,83 @@
...
+#include <cuda_runtime.h>
+
+#include <nvbench/cuda_call.cuh>
+#include <nvbench/cuda_stream.cuh>
+#include <nvbench/detail/throw.cuh>
```
alliepiper, line 24: unused.

nvbench/timestamps_kernel.cu
```diff
@@ -0,0 +1,83 @@
...
+
+void timestamps_kernel::record(const nvbench::cuda_stream &stream)
+{
+  m_host_timestamps[0] = 0;
+  m_host_timestamps[1] = 0;
+
+  int device_id = 0;
+  int num_sms   = 0;
+
+  NVBENCH_CUDA_CALL_NOEXCEPT(cudaGetDevice(&device_id));
```
alliepiper, line 75: minor: why use the noexcept version here? Let's use the throwing check or mark the function as `noexcept` if these are needed.

nvbench/benchmark_base.cuh
```diff
@@ -271,6 +298,10 @@ protected:
   nvbench::float64_t m_skip_time{-1.};
   nvbench::float64_t m_timeout{15.};
 
+  nvbench::float32_t m_throttle_threshold{0.8f};      // [% of peak SM clock rate]
+  nvbench::float32_t m_throttle_recovery_delay{0.0f}; // [seconds]
+  bool m_discard_on_throttle{false};
```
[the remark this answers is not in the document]
gevtushenko: Completely forgot about these.. Thanks, @alliepiper!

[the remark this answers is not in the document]
alliepiper: It's easy to do, I've forgotten about it a few times, too. I'd like to find a better pattern for this at some point to make it less manual.

nvbench/detail/measure_cold.cu
```diff
@@ -78,6 +85,32 @@ void measure_cold_base::run_trials_prologue() { m_walltime_timer.start(); }
 
 void measure_cold_base::record_measurements()
 {
+  auto peak_clock_rate = m_state.get_device()->get_sm_peak_clock_rate();
```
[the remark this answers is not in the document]
gevtushenko: Benchmarking at lower frequencies is bad practice. If someone does this and want to get rid of warnings, they can set the threshold to zero.

docs/cli_help.md
```diff
@@ -133,6 +133,23 @@
   * Applies to the most recent `--benchmark`, or all benchmarks if specified
     before any `--benchmark` arguments.
 
+* `--discard-on-throttle`
```
alliepiper, line 136: I think we should just make this the default behavior. I don't think we need an opt-out -- if throttle checks are enabled, always discard throttled results.

nvbench/benchmark_base.cuh
```diff
@@ -282,6 +309,10 @@ protected:
   nvbench::float64_t m_skip_time{-1.};
   nvbench::float64_t m_timeout{15.};
 
+  nvbench::float32_t m_throttle_threshold{0.75f};     // [% of peak SM clock rate]
+  nvbench::float32_t m_throttle_recovery_delay{0.0f}; // [seconds]
```
alliepiper, line 313: Important: This doesn't match the documentation.

nvbench/detail/measure_cold.cuh
```diff
@@ -97,6 +102,10 @@ protected:
   nvbench::float64_t m_skip_time{};
   nvbench::float64_t m_timeout{};
 
+  nvbench::float32_t m_throttle_threshold{0.75f};     // [% of peak SM clock rate]
```
alliepiper, line 105: minor: No need to provide default values here -- they're always set from the `state`. This way there's less to keep synced with the docs (delay currently doesn't match docs).

nvbench/detail/measure_cold.cu
```diff
@@ -298,6 +341,17 @@ void measure_cold_base::generate_summaries()
     summ.set_string("hide", "Hidden by default.");
   }
 
+  if (!m_sm_clock_rates.empty())
+  {
+    auto &summ = m_state.add_summary("nv/cold/sm_clock_rate/mean");
```
alliepiper, line 346: Suggestion: Having an additional hidden column with this as a percentage of peak would be useful, too.

nvbench/state.cuh
```diff
@@ -311,6 +338,10 @@ private:
   nvbench::float64_t m_skip_time;
   nvbench::float64_t m_timeout;
 
+  nvbench::float32_t m_throttle_threshold{0.75f};     // [% of peak SM clock rate]
+  nvbench::float32_t m_throttle_recovery_delay{0.0f}; // [seconds]
```
alliepiper, line 342: Doesn't match docs, can be removed.

docs/cli_help.md
```diff
@@ -133,6 +133,23 @@
...
+
+* `--throttle-threshold <value>`
+  * Set the GPU throttle threshold as percentage of the peak clock rate.
+  * Default is 0.75 (75%).
+  * Applies to the most recent `--benchmark`, or all benchmarks if specified
+    before any `--benchmark` arguments.
+
+* `--throttle-recovery-delay <value>`
+  * Set the GPU throttle recovery delay in seconds.
+  * Default is 0.5 seconds.
```
alliepiper, line 149: Suggestion: I think we could be more aggressive here. Just 0.05s was enough to get the frequency back up to a suitable rate for at least a few more samples in my testing (A6000, cub.select_if/cub.radix_sort), even with the threshold set to 90%.

My understanding is that as long as we're not constantly delaying for the same sample without making progress we can minimize this as much as possible.

alliepiper (commented):
I had some time to test this out on CUB's benchmarks this morning and left some feedback. I'll give this a more thorough look next week.

nvbench/detail/measure_cold.cu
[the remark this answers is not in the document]
alliepiper: > Benchmarking at lower frequencies is bad practice.

I'm still not completely convinced of that, but I think with features like this throttle detect/discard I'm starting to see the light 😄

alliepiper:
Watching the logs as these run gives some good insight into which configurations are really stressing the hardware. It's neat to see in real time.

docs/cli_help.md
```diff
@@ -133,6 +133,23 @@
...
     before any `--benchmark` arguments.
 
+* `--discard-on-throttle`
+  * Discard measurements if the GPU is throttled.
+  * Applies to the most recent `--benchmark`, or all benchmarks if specified
+    before any `--benchmark` arguments.
+
+* `--throttle-threshold <value>`
+  * Set the GPU throttle threshold as percentage of the peak clock rate.
+  * Default is 0.75 (75%).
```
alliepiper, line 143: important: Existing options (eg. `--max-noise`) take a percentage as a percentage, not a fraction. Let's change this to take 75 instead of 0.75 for consistency.

nvbench/detail/measure_cold.cu
gevtushenko: I'll have to leave this up to you to address as a follow up

nvbench/detail/measure_cold.cu
```diff
@@ -298,6 +341,17 @@ void measure_cold_base::generate_summaries()
...
+    summ.set_string("name", "Clock Rate");
+    summ.set_string("hint", "frequency");
+    summ.set_string("description", "Mean SM clock rate");
+    summ.set_string("hide", "Hidden by default.");
+    summ.set_float64("value", nvbench::detail::statistics::compute_mean(m_sm_clock_rates.cbegin(),
```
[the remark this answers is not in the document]
gevtushenko: I'll have to leave this up to you to address as a follow up

PR #212, opened 2025-04-15
Fix issue with newer thrust.

alliepiper:
Implicit conversions from `nvbench::cuda_stream` -> `cudaStream_t` no longer work when creating Thrust execution policies, as the `cudaStream_t` overload of `thrust::device.on(...)` was removed and replaced with `cuda::stream_ref`.

Explicitly pulling the `cudaStream_t` out of the `nvbench::cuda_stream` and passing it to `on` so that the implicit constructor of `stream_ref` can handle it seems to be the only way to fix this while preserving support back to CTK 12.0 (`stream_ref` was added in CTK 12.3).


alliepiper:
Waiting to see if we can get https://github.com/NVIDIA/cccl/pull/4451 instead.

alliepiper:
We fixed this by reverting the breaking change in Thrust instead.

PR #214, opened 2025-04-23
Fix throttle setter return values and update customization example

PointKernel:
This PR fixes a bug where the benchmark throttle setters were incorrectly returning void instead of the updated benchmark settings. It also updates the example to demonstrate how to customize the throttle threshold.

### DIFF
--- a/examples/custom_criterion.cu
+++ b/examples/custom_criterion.cu
@@ -74,4 +74,4 @@ void throughput_bench(nvbench::state &state)
       num_values);
   });
 }
-NVBENCH_BENCH(throughput_bench).set_stopping_criterion("fixed").set_throttle_threshold(0.8f);
+NVBENCH_BENCH(throughput_bench).set_stopping_criterion("fixed");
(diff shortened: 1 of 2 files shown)


examples/custom_criterion.cu
```diff
@@ -74,4 +74,4 @@ void throughput_bench(nvbench::state &state)
       num_values);
   });
 }
-NVBENCH_BENCH(throughput_bench).set_stopping_criterion("fixed");
+NVBENCH_BENCH(throughput_bench).set_stopping_criterion("fixed").set_throttle_threshold(0.8f);
```
PointKernel, line 77: Without the changes in this PR, this code will fail to compile.

alliepiper: The changes to the benchmark class are definitely necessary, but I'd rather not have this here. This example is showing how to modify the custom criterion, and the throttle settings aren't really related to that.

I'll be happy to merge this without this modification. If we want an example of this, we should add a new `throttle_detection` example with comments/annotations explaining the throttle options.

alliepiper (changes requested):

examples/custom_criterion.cu
PointKernel: Totally makes sense. Changes reverted.

alliepiper (approved):
LGTM, thanks for the fix!

Issue #247, opened 2025-07-17
Show batch time comparisons in `nvbench_compare.py`


bernhardmgruber:
When comparing benchmark results with `nvbench_compare.py`, currently, the cold measurements are compared and shown. If a JSON file also contains batch measurements, those should be shown in addition.

ahendriksen:
The batch timings are often ignored, see also #194 . Perhaps it is possible to introduce an abstraction that makes it easy to always grab the "best" timing, depending on whether batch or non-batch timing is available.

PR #263, opened 2025-08-14
Also compare batch measurements in nvbench_compare.py

bernhardmgruber:
Fixes: #247

### DIFF
--- a/scripts/nvbench_compare.py
+++ b/scripts/nvbench_compare.py
@@ -137,6 +137,8 @@ def compare_benches(ref_benches, cmp_benches, threshold, plot):
         headers.append("Status")
         colalign.append("center")
 
+        added_batch_headers = False
+
         for device_id in device_ids:
             rows = []
             plot_data = {"cmp": {}, "ref": {}, "cmp_noise": {}, "ref_noise": {}}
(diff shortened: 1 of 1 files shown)


python/scripts/nvbench_compare.py
```diff
@@ -269,6 +300,19 @@ def extract_value(summary):
                     failure_count += 1
                     status = Fore.RED + "SLOW" + Fore.RESET
 
+                if has_batch_data:
+                    if (
+                        abs(frac_diff_batch) <= 0.01
+                    ):  # TODO(bgruber): what value to use here?
```
alliepiper, line 312: I have no idea, let's get some input internally on that.

alliepiper: Maybe just pick a sensible default and let the user override with command-line opts?

oleksandr-pavlyk: We should use `min_noise` estimated from  `abs(frac_diff_batch) <= min_noise` if available, like we do for cold measurements, and use yellow tint if `min_noise` is not available. 

Perhaps in case when `has_batch_data`  is `True`, `min_noise` should always be available.

PR #269, opened 2025-09-09
remove pynvjitlink references in examples

jayavenkatesh19:
Towards https://github.com/rapidsai/build-planning/issues/210

Based on the information from [this comment on the pynvjitlink repo](https://github.com/rapidsai/pynvjitlink/pull/129#issuecomment-3004822494) and this [section of the numba-cuda docs](https://nvidia.github.io/numba-cuda/reference/kernel.html#kernel-declaration), it looks like both the linker and LTO are already configured as desired in the newer `numba-cuda` versions. 

So it should be fine to just remove the references to `pynvjitlink` without altering anything else

### DIFF
--- a/python/examples/cccl_cooperative_block_reduce.py
+++ b/python/examples/cccl_cooperative_block_reduce.py
@@ -21,7 +21,6 @@ import cuda.cccl.cooperative.experimental as coop
 import numba
 import numpy as np
 from numba import cuda
-from pynvjitlink import patch
 
 
 class BitsetRing:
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
@jayavenkatesh19 Should "requirements.txt" be modified to pin `numba-cuda` from below?

jayavenkatesh19:
Oh yes! Apologies for missing that!

Pushed a new commit that adds `numba-cuda` to the `requirements.txt`

PR #271, opened 2025-09-19
Add best practices documentation

PointKernel:
This PR adds a best practices guide for NVBench, providing code examples to help users quickly get started and conduct effective performance comparisons in real-world scenarios.

### DIFF
--- a/docs/best_practices.md
+++ b/docs/best_practices.md
@@ -127,16 +127,9 @@ Pass: Cold: 0.007819ms GPU, 0.013864ms CPU, 0.50s total GPU, 3.59s total wall, 6
 |  63952x | 13.864 us | 432.95% | 7.819 us | 447.95% |
 ```
 
-By default, NVBench runs benchmarks on all available GPUs unless specified otherwise. On multi-GPU systems, this can unnecessarily increase runtime and resource usage. To target a specific GPU, saving both time and resources, you can set the `CUDA_VISIBLE_DEVICES` environment variable. In our case, we target the **RTX8000**:
-
+By default, NVBench executes benchmarks on all available GPUs unless instructed otherwise. On multi-GPU systems, this can lead to longer runtimes and higher resource usage. To focus on a specific GPU and optimize both time and resources, users can use the `-d` CLI option to select the target GPU. In our example, we target the **RTX8000**:
 ```bash
-user@nvbench-test:~/nvbench/build/bin$ export CUDA_VISIBLE_DEVICES=0
-```
-
-Now, if we rerun:
-
-```bash
-user@nvbench-test:~/nvbench/build/bin$ ./sequence_bench
+user@nvbench-test:~/nvbench/build/bin$ ./sequence_bench -d 0
 # Devices
 
 ## [0] `Quadro RTX 8000`


docs/best_practices.md
```diff
@@ -0,0 +1,297 @@
...
+```cpp
+void sequence_bench(nvbench::state& state) {
+  auto data = thrust::device_vector<int>(10);
+  state.exec([](nvbench::launch& launch) {
+    thrust::sequence(data.begin(), data.end());
+  });
+}
+NVBENCH_BENCH(sequence_bench);
+```
+Will this code work as-is? Depending on the build system configuration, compilation may succeed but generate warnings indicating that `launch` is an unused parameter. The code may or may not execute correctly. This often occurs when users, accustomed to a sequential programming mindset, overlook the fact that GPU architectures are highly parallel. Proper use of streams and synchronization is essential for accurately measuring performance in benchmark code.
```
oleksandr-pavlyk: Perhaps the sentence about possible warning is superfluous. We can resolve the warning by simply removing the unused name, i.e., changing  `[](nvbench::launch& launch) {` to `[](nvbench::launch&) {`.

docs/best_practices.md
```diff
@@ -0,0 +1,297 @@
...
+|  81312x | 9.768 us | 13.55% | 6.150 us | 20.16% |
+
+### [1] NVIDIA RTX A400
+
+| Samples | CPU Time  |  Noise  | GPU Time |  Noise  |
+|---------|-----------|---------|----------|---------|
+|  63952x | 13.864 us | 432.95% | 7.819 us | 447.95% |
+```
+
+By default, NVBench runs benchmarks on all available GPUs unless specified otherwise. On multi-GPU systems, this can unnecessarily increase runtime and resource usage. To target a specific GPU, saving both time and resources, you can set the `CUDA_VISIBLE_DEVICES` environment variable. In our case, we target the **RTX8000**:
```
oleksandr-pavlyk: I think one should also mention that benchmark code executable supports `-d device_id` option, to only run benchmark on the specified device. So as an alternative to setting `CUDA_VISIBLE_DEVICES`, we can run `./sequence_bench -d 0` to only run benchmarks on device 0.

PointKernel: nvbench had multi-GPU issues with the `-d` flag for a long time. By default, my approach has always been to set the visible device before using nvbench. After looking further into the related issue, I noticed that https://github.com/NVIDIA/nvbench/pull/113 was merged about six months ago, finally resolving this problem. I’ll update the documentation to recommend using `-d` as the ~~primary~~ solution.~~, with setting the visible device as a backup alternative.~~


_Edit: probably not worth mentioning the visible device way._

PointKernel:
@oleksandr-pavlyk Thanks for the comments. Totally makes sense to me. I’ve mainly been gathering the caveats and pitfalls I paid close attention to when applying nvbench in cuCollections or cuDF, but I wasn’t sure which details are actually needed or appropriate for nvbench best practices. I’d be happy to go through your bullet points and contribute whatever comes to my mind, and then we can refine things from there. 

Regarding the comparison between nvbench and NCU, could you elaborate a bit on what kind of information would be most useful for users? From my perspective, nvbench primarily gathers runtime data and basic metrics, which feels more similar to NSYS, whereas NCU provides deeper kernel-level profiling with detailed hardware utilization insights.

PR #277, opened 2025-10-08
Profile only the kernels involved in the benchmark

comeyrd:


comeyrd:
pre-commit.ci autofix

nvbench/detail/measure_cold.cuh
```diff
@@ -152,6 +158,10 @@ struct measure_cold_base::kernel_launch_timer
     {
       m_measure.gpu_frequency_start();
     }
+    if (!m_run_once)
```
alliepiper: The logic here is reversed -- we only want to execute this when `run_once` is true (eg. when profiling).

```suggestion
    if (m_run_once)
```

nvbench/detail/measure_cold.cuh
```diff
@@ -170,12 +180,17 @@ struct measure_cold_base::kernel_launch_timer
       m_measure.gpu_frequency_stop();
     }
     m_measure.sync_stream();
+    if (!m_run_once)
```
alliepiper:
```suggestion
    if (m_run_once)
```

Issue #121, opened 2023-01-25
Custom CUPTI metrics


mmcloughlin:
We would like to support additional CUPTI metric types. As I understand the code right now, it seems there is a fixed set of supported metrics. As noted in #91 it's not that hard to submit a PR that adds another one, but we'd like to be able to extend this set without having to either submit a PR to nvbench, or maintain our own fork.

Would it be possible to provide an interface that can enable an arbitrary custom metric?

saurabh-singh-rajput:
Hi there, I was looking for the same thing but couldn't see much update. Is there a way to do this?

PR #283, opened 2025-10-22
Adding the possibility of having user selected Cupti metrics

comeyrd:
This pull request is based on the idea from the issue [#121](https://github.com/NVIDIA/nvbench/issues/121).

The user would be able to provide Cupti metrics it wants to use for his benchmark.
By filling one or multiple CustomCuptiMetrics
```cpp
struct CustomCuptiMetrics{
    const char *metric_name;
    const char *name;
    const char *hint;
    const char *description;
    const double divider;
  }
```
The ``measure_cupti.cuh/cu`` would add these custom metrics to the nvbench defined ones, forwarding them to ``cupti_profiler.cuh/cxx``.
To support that, ``cupti_profiler.cuh/cxx`` would need to check if the selected metrics are available on the device, dropping the unavailable metrics and running the experiment even if some metrics are not available.

The ``measure_cupti.cuh/cu`` would handle displaying that metrics were unavailable.

The availability of a metric would be queried using [Perfworks Metric API](https://docs.nvidia.com/cupti/main/main.html#perfworks-metric-api) and the ``NVPW_MetricsEvaluator_GetMetricNames()`` API call to see if the user-specified metrics are available on the current device.

For now only some comments for understanding the actual implementation are on this branch, and also places where I think I will add the required code.

Any thoughts on this Project ?

### DIFF
--- a/nvbench/cupti_profiler.cxx
+++ b/nvbench/cupti_profiler.cxx
@@ -68,6 +68,7 @@ cupti_profiler::cupti_profiler(nvbench::device_info device, std::vector<std::str
   initialize_profiler();
   initialize_chip_name();
   initialize_availability_image();
+  verify_metric_names();
   /*
     Update the std::vector<std::string> metric_names with the metrics that are available within the GPU.
     Failing gracefully will enable the measurement of other metrics if a metric is not available
@@ -146,14 +147,6 @@ void cupti_profiler::initialize_availability_image()
   cupti_call(cuptiProfilerGetCounterAvailability(&params));
 }
 
-void cupti_profiler::verify_metric_names()
-{
-  metric_evaluator evaluator(m_chip_name, m_availability_image.data());
-  m_verified_metric_names = evaluator.list_metrics();
-  for(std::string item:m_verified_metric_names){
-    printf("%s\n", item.c_str());//TOCHECK
-  }
-}
 
 void cupti_profiler::initialize_nvpw()
 {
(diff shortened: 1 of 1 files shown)


comeyrd:
Hi @gevtushenko  !
I am working on the ``cupti_profiler`` that you wrote a few years ago, 
I just have one question about the ``counter_data_builder`` and it's use.
in cupti_profiler.cxx, line 476 (or around that) 
```
void cupti_profiler::initialize_counter_data_prefix_image()
{
  const std::uint8_t *counter_availability_image = nullptr;

  std::vector<NVPA_RawMetricRequest> raw_metric_requests =
    get_raw_metric_requests(m_chip_name, m_metric_names, counter_availability_image);

  counter_data_builder data_builder(m_chip_name, counter_availability_image);
...
```

For what I understood, the availability image is an array of uint8_t, that will be used in the futur (or is used now) instead of the chip_name to identify the chip and its capabilities.

in cupti_profiler.cxx, line 133 (or around that) you initialise the member variable ``m_availability_image``, 
```
void cupti_profiler::initialize_availability_image()
{
  CUpti_Profiler_GetCounterAvailability_Params params{};

  params.structSize = CUpti_Profiler_GetCounterAvailability_Params_STRUCT_SIZE;
  params.ctx        = m_device.get_context();

  cupti_call(cuptiProfilerGetCounterAvailability(&params));

  m_availability_image.clear();
  m_availability_image.resize(params.counterAvailabilityImageSize);
  params.pCounterAvailabilityImage = m_availability_image.data();

  cupti_call(cuptiProfilerGetCounterAvailability(&params));
}
```

But in ``cupti::initialize_counter_data_prefix_image()`` you give the ``counter_data_builder`` a nullptr availability image.

Can you give me insight on that ? 
Is there other documentation about cupti else than [Cupti Usage](https://docs.nvidia.com/cupti/main/main.html#usage), is there an API reference ? Or the only info I can access is in the different headers of the library ?

Thanks !

PR #286, opened 2025-11-21
entropy criterion optimizations

jhou-jpg:
**Overview**
This PR significantly optimizes the entropy criterion algorithm by reducing its computational complexity from O(n) to O(1) per update. The new approach leverages incremental updates and efficient data structures to avoid repeated full traversals, resulting in substantial performance improvements, especially for large datastreams.

**Performance**
Initial benchmarks demonstrate a significant improvement in update speed (see attached image for details).
The optimized implementation consistently matches the entropy results of the current version across a comprehensive suite of tests.

Operation | Previous Time Complexity | New Time Complexity
-- | -- | --
Add Element | O(n) | O(1)
Compute Entropy | O(n) | O(1)
Is Finished Check | O(n) | O(1)

**Test Plan**
The current and new implementation were recreated in a sandbox environment to run and test on Meta's buck build system.

image

image

image

### DIFF
--- a/nvbench/detail/online_linear_regression.cuh
+++ b/nvbench/detail/online_linear_regression.cuh
@@ -104,7 +104,7 @@ public:
 
   [[nodiscard]] nvbench::float64_t slope() const
   {
-    static constexpr nvbench::float_64_t q_nan =
+    static constexpr nvbench::float64_t q_nan =
       std::numeric_limits<nvbench::float64_t>::quiet_NaN();
 
     if (m_count < 2)
@@ -152,7 +152,7 @@ public:
     const nvbench::float64_t mean_y_v = mean_y();
     const nvbench::float64_t ss_tot   = (m_sum_y2 / n) - mean_y_v * mean_y_v;
 
-    if (ss_tot == 0)
+    if (ss_tot < std::numeric_limits<nvbench::float64_t>::epsilon())
     {
       return 1.0;
     }
@@ -166,12 +166,15 @@ public:
     }
     else
     {
-      const nvbench::float64_t ss_res =
-        (m_sum_y2 / n) - 2.0 * slope_v * (m_sum_xy / n) - 2.0 * intercept_v * (m_sum_y / n) +
-        slope_v * slope_v * (m_sum_x2 / n) + 2.0 * slope_v * intercept_v * (m_sum_x / n) +
-        intercept_v * intercept_v;
+      const nvbench::float64_t mean_xy_v = m_sum_xy / n;
+      const nvbench::float64_t mean_xx_v = m_sum_x2 / n;
+      const nvbench::float64_t mean_x_v  = m_sum_x / n;
+      const nvbench::float64_t ss_tot_m_res =
+        slope_v * ((mean_xy_v - slope_v * mean_xx_v) + (mean_xy_v - intercept_v * mean_x_v)) +
+        intercept_v * (mean_y_v - slope_v * mean_x_v - intercept_v) +
+        mean_y_v * (intercept_v - mean_y_v);
 
-      return std::max(0.0, std::min(1.0, 1.0 - (ss_res / ss_tot)));
+      return std::min(std::max(ss_tot_m_res / ss_tot, 0.0), 1.0);
     }
   }
 


adamomainz:
LGTM! Lets wait for the maintainers to approve or leave comments!

nvbench/detail/entropy_criterion.cxx
```diff
@@ -37,37 +36,50 @@ void entropy_criterion::do_initialize()
...
   {
-    entropy -= p * std::log2(p);
+    return 0.0;
   }
 
-  return entropy;
+  const nvbench::float64_t n_float = static_cast<nvbench::float64_t>(n);
+  const nvbench::float64_t entropy = std::log2(n_float) - m_sum_count_log_counter / n_float;
+
+  return (std::max(0.0, entropy) == 0.0) ? 0.0 : entropy;
```
oleksandr-pavlyk: Suggestion: replace with `return std::max(0.0, entropy);`

nvbench/detail/entropy_criterion.cxx
```diff
@@ -37,37 +36,50 @@ void entropy_criterion::do_initialize()
...
-  for (nvbench::float64_t p : m_probabilities)
+  if (n == 0)
   {
-    entropy -= p * std::log2(p);
+    return 0.0;
   }
 
-  return entropy;
+  const nvbench::float64_t n_float = static_cast<nvbench::float64_t>(n);
+  const nvbench::float64_t entropy = std::log2(n_float) - m_sum_count_log_counter / n_float;
```
oleksandr-pavlyk:
```suggestion
  if (m_total_samples == 0)
{
    return 0.0;
}

  const auto n = static_cast<nvbench::float64_t>(m_total_samples);
  const nvbench::float64_t entropy = std::log2(n) - m_sum_count_log_counter / n;
```

nvbench/detail/entropy_criterion.cxx
jhou-jpg: the current implementation prevents -0.0000 since c++ see -0.00000 == 0.00000

nvbench/detail/entropy_criterion.cxx
```diff
@@ -88,15 +100,53 @@ void entropy_criterion::do_add_measurement(nvbench::float64_t measurement)
...
     else
     {
+      old_count = 0;
       m_freq_tracker.insert(it, std::make_pair(key, nvbench::int64_t{1}));
+      new_count = 1;
     }
   }
 
-  m_entropy_tracker.push_back(compute_entropy());
+  update_entropy_sum(old_count, new_count);
```
oleksandr-pavlyk: Note that `new_count` is always `old_count + 1`, so perhaps there is no need to introduce `new_count` variable.

nvbench/detail/entropy_criterion.cxx
```diff
@@ -37,37 +36,50 @@ void entropy_criterion::do_initialize()
...
   }
 
-  m_probabilities.resize(n);
-  for (std::size_t i = 0; i < n; i++)
+  if (new_count > 0)
   {
-    m_probabilities[i] = static_cast<nvbench::float64_t>(m_freq_tracker[i].second) /
-                         static_cast<nvbench::float64_t>(m_total_samples);
+    m_sum_count_log_counter += new_count * std::log2(new_count);
   }
```
oleksandr-pavlyk, line 56:
```suggestion
   assert(new_count > old_count);
   if (old_count  > 0) {
       auto diff = new_count - old_count;
       m_sum_count_log_counter += new_count * std::log1p(diff/old_count) + diff * std::log(old_count);
   } else {
       m_sum_count_log_counter +=new_count *  std::log(new_count);
   }
```

For $o > 0$ we have

```math
 n \cdot \log(n) - o \cdot  \log(o) =
    n \cdot  \log\left(\frac{n}{o}\right) + n \cdot  \log(o) - o \cdot  \log(o) =
    n \cdot  \log\left(1 + \frac{n-o}{o}\right) + (n - o) \cdot  \log(o)
```

The advantage of this reformulation is that it is more accurate when `n` and `o` are large, and `n-o` is small.

jhou-jpg: wow, very cool!

nvbench/detail/entropy_criterion.cxx
```diff
@@ -88,15 +100,53 @@ void entropy_criterion::do_add_measurement(nvbench::float64_t measurement)
...
+  const nvbench::float64_t entropy = compute_entropy();
+  const nvbench::float64_t n       = static_cast<nvbench::float64_t>(m_entropy_tracker.size() - 1);
+
+  // remove oldest value from stats if buffer is full
+  if (m_entropy_tracker.size() == m_entropy_tracker.capacity())
+  {
+    // Buffer is full, need to remove oldest entropy value
+    const nvbench::float64_t old_entropy = *m_entropy_tracker.cbegin();
+    const nvbench::float64_t old_x       = 0.0; // Oldest position in sliding window
+    const nvbench::float64_t old_sum_x   = m_sum_x;
```
oleksandr-pavlyk: `old_sum_x` is unused, please remove.

nvbench/detail/entropy_criterion.cxx
```diff
@@ -88,15 +100,53 @@ void entropy_criterion::do_add_measurement(nvbench::float64_t measurement)
...
+
+    // Remove old value contributions
+    m_sum_x -= old_x + n;
+    m_sum_y -= old_entropy;
+    m_sum_xy -= old_x * old_entropy;
+    m_sum_x2 -= old_x * old_x;
+    m_sum_y2 -= old_entropy * old_entropy;
```
oleksandr-pavlyk: Since `old_x = 0.0` in this scope, please simplify to remove unnecessary code

nvbench/detail/entropy_criterion.cxx
oleksandr-pavlyk: Thanks for the context. In this case, I'd suggest `return std::copysign(std::max(0.0, entropy), 1.0);` instead. 

See https://godbolt.org/z/Tv6EWs1rT to inspect assembly generated.

<details>
<summary>Code used for timing</summary>

```cpp
// g++ a.cpp 
#include <iostream>
#include <cmath>
#include <chrono>
#include <algorithm>
#include <vector>

double foo_1(double x) {
    return std::abs(std::max(0.0, x));
}

double foo_2(double x) {
    return std::copysign(std::max(0.0, x), 1.0);
}

double foo_3(double x) {
    return (std::max(0.0, x) == 0.0) ? 0.0 : x;
}

int main(void) {
   std::size_t n = 16*1024;
   std::vector<double> inp(n, -0.0);
   std::vector<double> out{};
   out.reserve(n);

   size_t n_reps = 16*1024;

   auto t0 = std::chrono::steady_clock::now();
   for(size_t r = 0; r < n_reps; ++r) {
      std::transform(inp.begin(), inp.end(), out.begin(), foo_1);
   }
   auto t1 = std::chrono::steady_clock::now();

   for(size_t r = 0; r < n_reps; ++r) {
      std::transform(inp.begin(), inp.end(), out.begin(), foo_2);
   }

   auto t2 = std::chrono::steady_clock::now();

   for(size_t r = 0; r < n_reps; ++r) {
      std::transform(inp.begin(), inp.end(), out.begin(), foo_3);
   }

   auto t3 = std::chrono::steady_clock::now();

   auto av_dur_nanoseconds = [n_reps](auto start, auto stop) -> double {
      return std::chrono::duration_cast<std::chrono::nanoseconds>(stop - start).count() / static_cast<double>(n_reps);
   };

   std::cout << "foo_1: " << av_dur_nanoseconds(t0, t1)<< "\n";
   std::cout << "foo_2: " << av_dur_nanoseconds(t1, t2) << "\n";
   std::cout << "foo_3: " << av_dur_nanoseconds(t2, t3) << "\n";

   return 0;
}
```
</details>

The results of timing (in nanoseconds per repetition) observed in my WSL box on Core Ultra 7 laptop:

```
foo_1: 322895
foo_2: 306901
foo_3: 324151
```

This suggests that the suggested edit would result in slightly faster code.

nvbench/detail/entropy_criterion.cuh
```diff
@@ -33,14 +33,20 @@ class entropy_criterion final : public stopping_criterion_base
...
   // TODO The window size should be user-configurable
   nvbench::detail::ring_buffer<nvbench::float64_t> m_entropy_tracker{299};
 
-  // Used to avoid re-allocating temporary memory
-  std::vector<nvbench::float64_t> m_probabilities;
+  nvbench::float64_t m_sum_x{};
+  nvbench::float64_t m_sum_y{};
+  nvbench::float64_t m_sum_xy{};
+  nvbench::float64_t m_sum_x2{};
+  nvbench::float64_t m_sum_y2{};
```
oleksandr-pavlyk: These variables are associated with online computation of linear regression. Instead of introducing them here, could you introduce a class for computing linear regression with online update methods used in this PR?

The class would have `void update( std::pair<double, double> outgoing, std::pair<double, double> incoming)` where pair represents `(x, y)` and `void update( std::pair<double, double> incoming);` methods

nvbench/detail/entropy_criterion.cxx
oleksandr-pavlyk: Actually, since `x` variable values in the input data has values `0, 1, 2, ..., n - 1`, we know that `m_sum_x` value to stay the same at `n*(n-1)/2`, and `m_sum_x2` to stay the same at `n*(n-1)*(2*n-1)/6)`. 

See https://www.wolframalpha.com/input?i=sum%28k%5E2%2C+0%3C%3Dk%3Cn%29

This means that `m_sum_x` and `m_sum_x2` need not be recomputed at all.

jhou-jpg: yes that is right, the it should be a fixed number at the end, some of these were left here for visibility, refactoring

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,169 @@
+/*
+ *  Copyright 2023 NVIDIA Corporation
```
oleksandr-pavlyk: Perhaps 2025, as this is a new file.

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,169 @@
...
+  [[nodiscard]] nvbench::float64_t mean_y() const
+  {
+    return m_count > 0 ? m_sum_y / static_cast<nvbench::float64_t>(m_count) : 0.0;
+  }
+
+  [[nodiscard]] nvbench::float64_t slope() const
+  {
+    if (m_count < 2)
+    {
+      return 0.0;
```
oleksandr-pavlyk: Previously, for small `m_count` we would return `false` from the `is_finished` method.

Returning `0.0` slope does not amount to the same. I'd suggest returning `std::numeric_limits<nvbench::float64_t>::quiet_NaN()` (see [ref](https://en.cppreference.com/w/cpp/types/numeric_limits.html)) instead, and at the call site return `false` is the `std::isfinite(slope)` is not true.

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,169 @@
...
+    const nvbench::float64_t n      = static_cast<nvbench::float64_t>(m_count);
+    const nvbench::float64_t mean_x = this->mean_x();
+    const nvbench::float64_t mean_y = this->mean_y();
+
+    const nvbench::float64_t numerator   = m_sum_xy - n * mean_x * mean_y;
+    const nvbench::float64_t denominator = m_sum_x2 - n * mean_x * mean_x;
+
+    if (std::abs(denominator) < 1e-9)
+    {
+      return 0.0;
```
oleksandr-pavlyk: Same comment here

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    return numerator / denominator;
+  }
+
+  [[nodiscard]] nvbench::float64_t intercept() const
+  {
+    if (m_count < 2)
+    {
+      return 0.0;
```
oleksandr-pavlyk: same comment applies here

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    return mean_y() - slope() * mean_x();
+  }
+
+  [[nodiscard]] nvbench::float64_t r_squared() const
+  {
+    if (m_count < 2)
+    {
+      return 0.0;
```
oleksandr-pavlyk: ditto

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    const nvbench::float64_t n        = static_cast<nvbench::float64_t>(m_count);
+    const nvbench::float64_t mean_y_v = mean_y();
+    const nvbench::float64_t ss_tot   = m_sum_y2 - n * mean_y_v * mean_y_v;
+
+    if (ss_tot < 1e-9)
+    {
+      return 1.0;
```
oleksandr-pavlyk, line 157: Perhaps the same here.

nvbench/detail/online_linear_regression.cuh
jhou-jpg: thank you, great catch!

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,185 @@
...
+    {
+      return std::numeric_limits<nvbench::float64_t>::quiet_NaN();
+    }
+
+    const nvbench::float64_t n      = static_cast<nvbench::float64_t>(m_count);
+    const nvbench::float64_t mean_x = this->mean_x();
+    const nvbench::float64_t mean_y = this->mean_y();
+
+    const nvbench::float64_t numerator   = m_sum_xy - n * mean_x * mean_y;
+    const nvbench::float64_t denominator = m_sum_x2 - n * mean_x * mean_x;
```
oleksandr-pavlyk: The `n * mean_x` is just `m_sum_x`, so we can save one division as well as one conditional execution. 

We could also just do `numerator   = n * m_sum_xy - m_sum_x * m_sum_y;` and `denominator = n * m_sum_x2 - m_sum_x * m_sum_x;` and avoid calling `mean_x()` and `mean_y()` methods altogether.

If overflow is a concern, we should form `mean_xy` and `mean_x2` and use them to compute numerator and denominator. 

```cpp
   static constexpr nvbench::float_64_t q_nan = std::numeric_limits<nvbench::float64_t>::quiet_NaN();

   if (m_count == 0)  return q_nan;
   
   const nvbench::float64_t n      = static_cast<nvbench::float64_t>(m_count);
   const nvbench::float64_t mean_x = (m_sum_x / n);
   const nvbench::float64_t mean_y = (m_sum_y / n);
   
   const nvbench::float64_t numerator   = (m_sum_xy / n) - mean_x * mean_y;
   const nvbench::float64_t numerator   = (m_sum_x2 / n) - mean_x * mean_y;
   
   if (std::abs(denominator) < 1e-9) return q_nan;
   
   return numerator / denominator;
```

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,185 @@
...
+  [[nodiscard]] nvbench::float64_t r_squared() const
+  {
+    if (m_count < 2)
+    {
+      return std::numeric_limits<nvbench::float64_t>::quiet_NaN();
+    }
+
+    const nvbench::float64_t n        = static_cast<nvbench::float64_t>(m_count);
+    const nvbench::float64_t mean_y_v = mean_y();
+    const nvbench::float64_t ss_tot   = m_sum_y2 - n * mean_y_v * mean_y_v;
```
oleksandr-pavlyk: Minor suggestion: perhaps also scale both `ss_tot` and `ss_res` by dividing each by `n` to concerns of hitting an overflow.

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,185 @@
...
+
+    if (ss_tot < 1e-9)
+    {
+      return 1.0;
```
oleksandr-pavlyk, line 157: **Q**: could you explain why it makes sense to return `1.0` (perfect fit) here?  If `ss_tot` is "almost zero", forming `(ss_res / ss_tot)` is "not meaningful".

oleksandr-pavlyk:
After build break is addressed, the test suite runs clean. 

Comparisons of test output of `./bin/nvbench.example.cpp20.axes -d 0 -b copy_type_and_block_size_sweep` executed from this branch's and the main branch's build directories:

```
$ scripts/nvbench_compare.py review.json main.json
['review.json', 'main.json']
# copy_type_and_block_size_sweep

## [0] NVIDIA RTX A6000

|  Type  |  BlockSize  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |       Diff |   %Diff |  Status  |
|--------|-------------|------------|-------------|------------|-------------|------------|---------|----------|
|   I8   |     64      |   2.668 ms |       3.80% |   2.655 ms |       2.62% | -12.583 us |  -0.47% |   SAME   |
|   I8   |     128     |   1.321 ms |       0.38% |   1.303 ms |       0.04% | -18.345 us |  -1.39% |   FAST   |
|   I8   |     196     |   1.478 ms |       0.11% |   1.465 ms |       0.11% | -13.247 us |  -0.90% |   FAST   |
|   I8   |     256     |   1.184 ms |       0.15% |   1.175 ms |       0.16% |  -8.757 us |  -0.74% |   FAST   |
|   I8   |     320     |   1.376 ms |       0.12% |   1.370 ms |       0.12% |  -6.881 us |  -0.50% |   FAST   |
|   I8   |     512     |   1.196 ms |       0.18% |   1.192 ms |       0.17% |  -3.378 us |  -0.28% |   FAST   |
|  I16   |     64      |   1.322 ms |       0.03% |   1.310 ms |       0.03% | -12.394 us |  -0.94% |   FAST   |
|  I16   |     128     | 814.080 us |       0.50% | 813.157 us |       0.50% |  -0.923 us |  -0.11% |   SAME   |
|  I16   |     196     | 858.350 us |       0.32% | 859.287 us |       0.25% |   0.937 us |   0.11% |   SAME   |
|  I16   |     256     | 813.381 us |       0.50% | 813.482 us |       0.49% |   0.102 us |   0.01% |   SAME   |
|  I16   |     320     | 838.446 us |       0.45% | 839.068 us |       0.44% |   0.622 us |   0.07% |   SAME   |
|  I16   |     512     | 816.853 us |       0.50% | 816.614 us |       0.50% |  -0.239 us |  -0.03% |   SAME   |
|  I32   |     64      | 782.066 us |       0.50% | 782.516 us |       0.50% |   0.450 us |   0.06% |   SAME   |
|  I32   |     128     | 787.193 us |       0.50% | 786.336 us |       0.50% |  -0.856 us |  -0.11% |   SAME   |
|  I32   |     196     | 780.827 us |       0.50% | 780.840 us |       0.50% |   0.012 us |   0.00% |   SAME   |
|  I32   |     256     | 788.869 us |       0.50% | 787.764 us |       0.50% |  -1.105 us |  -0.14% |   SAME   |
|  I32   |     320     | 784.653 us |       0.50% | 784.496 us |       0.50% |  -0.157 us |  -0.02% |   SAME   |
|  I32   |     512     | 786.413 us |       0.50% | 785.031 us |       0.50% |  -1.382 us |  -0.18% |   SAME   |
|  F32   |     64      | 782.631 us |       0.50% | 782.996 us |       0.50% |   0.365 us |   0.05% |   SAME   |
|  F32   |     128     | 787.208 us |       0.50% | 786.391 us |       0.50% |  -0.817 us |  -0.10% |   SAME   |
|  F32   |     196     | 780.803 us |       0.50% | 780.955 us |       0.50% |   0.152 us |   0.02% |   SAME   |
|  F32   |     256     | 788.809 us |       0.50% | 787.761 us |       0.50% |  -1.048 us |  -0.13% |   SAME   |
|  F32   |     320     | 784.694 us |       0.50% | 784.239 us |       0.50% |  -0.455 us |  -0.06% |   SAME   |
|  F32   |     512     | 786.664 us |       0.50% | 784.893 us |       0.50% |  -1.771 us |  -0.23% |   SAME   |
|  I64   |     64      | 785.511 us |       0.50% | 785.082 us |       0.50% |  -0.429 us |  -0.05% |   SAME   |
|  I64   |     128     | 791.982 us |       0.50% | 792.070 us |       0.50% |   0.088 us |   0.01% |   SAME   |
|  I64   |     196     | 786.148 us |       0.50% | 786.042 us |       0.50% |  -0.106 us |  -0.01% |   SAME   |
|  I64   |     256     | 791.823 us |       0.50% | 791.818 us |       0.50% |  -0.004 us |  -0.00% |   SAME   |
|  I64   |     320     | 788.731 us |       0.50% | 788.361 us |       0.50% |  -0.370 us |  -0.05% |   SAME   |
|  I64   |     512     | 791.668 us |       0.50% | 791.398 us |       0.50% |  -0.270 us |  -0.03% |   SAME   |
|  F64   |     64      | 785.479 us |       0.50% | 785.056 us |       0.50% |  -0.423 us |  -0.05% |   SAME   |
|  F64   |     128     | 791.929 us |       0.50% | 792.050 us |       0.50% |   0.121 us |   0.02% |   SAME   |
|  F64   |     196     | 786.136 us |       0.50% | 785.976 us |       0.50% |  -0.160 us |  -0.02% |   SAME   |
|  F64   |     256     | 791.870 us |       0.50% | 791.798 us |       0.50% |  -0.072 us |  -0.01% |   SAME   |
|  F64   |     320     | 788.691 us |       0.50% | 788.278 us |       0.50% |  -0.413 us |  -0.05% |   SAME   |
|  F64   |     512     | 791.634 us |       0.50% | 791.397 us |       0.50% |  -0.238 us |  -0.03% |   SAME   |

# Summary

- Total Matches: 36
  - Pass    (diff <= min_noise): 30
  - Unknown (infinite noise):    0
  - Failure (diff > min_noise):  6
```

The GPU kernel benchmarked did not change. Run-to-run variability is present in main as well.

Using a benchmark from the branch build directory, benchmark executable behavior with changes to entropy stopping criterion parameter changed is as expected:

```
nvbench) opavlyk@ee09c48-lcedt:~/repos/nvbench/build-review$ ./bin/nvbench.example.cpp20.axes -q -d 0 -b copy_type_and_block_size_sweep -a "Type=I32" -a "BlockSize=256" --stopping-criterion entropy --min-r2 0.93 --max-angle 0.0001 --markdown a_93.md
(nvbench) opavlyk@ee09c48-lcedt:~/repos/nvbench/build-review$ ./bin/nvbench.example.cpp20.axes -q -d 0 -b copy_type_and_block_size_sweep -a "Type=I32" -a "BlockSize=256" --stopping-criterion entropy --min-r2 0.95 --max-angle 0.0001 --markdown a_95.md

(nvbench) opavlyk@ee09c48-lcedt:~/repos/nvbench/build-review$ tail -n 7 a_93.md 
## copy_type_and_block_size_sweep

### [0] NVIDIA RTX A6000

| Type | BlockSize | ElementCount |    Input    |   Output    | Samples |  CPU Time  | Noise |  GPU Time  | Noise | Elem/s  | GlobalMem BW | BWUtil | Samples | Batch GPU  |
|------|-----------|--------------|-------------|-------------|---------|------------|-------|------------|-------|---------|--------------|--------|---------|------------|
|  I32 |       256 |     67108864 | 256.000 MiB | 256.000 MiB |   1436x | 798.682 us | 1.28% | 786.749 us | 0.21% | 85.299G | 682.392 GB/s | 88.84% |   1437x | 787.580 us |

(nvbench) opavlyk@ee09c48-lcedt:~/repos/nvbench/build-review$ tail -n 7 a_95.md 
## copy_type_and_block_size_sweep

### [0] NVIDIA RTX A6000

| Type | BlockSize | ElementCount |    Input    |   Output    | Samples |  CPU Time  | Noise |  GPU Time  | Noise | Elem/s  | GlobalMem BW | BWUtil | Samples | Batch GPU  |
|------|-----------|--------------|-------------|-------------|---------|------------|-------|------------|-------|---------|--------------|--------|---------|------------|
|  I32 |       256 |     67108864 | 256.000 MiB | 256.000 MiB |  10300x | 798.466 us | 0.86% | 786.833 us | 0.21% | 85.290G | 682.318 GB/s | 88.83% |  10301x | 787.304 us |

```

nvbench/detail/online_linear_regression.cuh
jhou-jpg: agree that it is not exactly meaningful, this was done to to return the "correct" r-value for constant value stream (eg. 5, 5, 5, 5)

jhou-jpg: https://github.com/NVIDIA/nvbench/blob/34f1e2a7eeeeec7bb2965a3ec9667d297b479cfe/nvbench/detail/statistics.cuh#L142

looks like statistics::compute_r2 before it here handled it similarly

jhou-jpg: kept the return 0 logic and lowered the tolerance from 1e-9 to 1e-12, happy to revise in a follwup issue

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  [[nodiscard]] nvbench::float64_t mean_y() const
+  {
+    return m_count > 0 ? m_sum_y / static_cast<nvbench::float64_t>(m_count) : 0.0;
+  }
+
+  [[nodiscard]] nvbench::float64_t slope() const
+  {
+    static constexpr nvbench::float_64_t q_nan =
```
oleksandr-pavlyk:
```suggestion
    static constexpr nvbench::float64_t q_nan =
```

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,185 @@
...
+      return std::numeric_limits<nvbench::float64_t>::quiet_NaN();
+    }
+    else
+    {
+      const nvbench::float64_t ss_res =
+        (m_sum_y2 / n) - 2.0 * slope_v * (m_sum_xy / n) - 2.0 * intercept_v * (m_sum_y / n) +
+        slope_v * slope_v * (m_sum_x2 / n) + 2.0 * slope_v * intercept_v * (m_sum_x / n) +
+        intercept_v * intercept_v;
+
+      return std::max(0.0, std::min(1.0, 1.0 - (ss_res / ss_tot)));
```
oleksandr-pavlyk: Evaluating `ss_tot - ss_res`  the term  `(m_sum_y2 / n)` cancels out. 

Since it can be large, I would propose to change this block to evaluate `ss_tot_m_res = ss_tot - ss_res` directly. One can also group terms in this difference which are likely to be small within the group.

```suggestion
      const nvbench::float64_t mean_xy_v = m_sum_xy / n;
      const nvbench::float64_t mean_xx_v = m_sum_x2 / n;
      const nvbench::float64_t mean_x_v = m_sum_x / n;
      const nvbench::float64_t ss_tot_m_res = 
        slope_v * (
            (mean_xy_v - slope_v * mean_xx_v) +
            (mean_xy_v - intercept_v * mean_x_v)
        ) + intercept_v * (mean_y_v - slope_v * mean_x_v - intercept_v) +
        mean_y_v * (intercept_v - mean_y_v);

      return std::min(std::max(ss_tot_m_res / ss_tot, 0.0), 1.0);
```

Here is proof of equivalency:

```python
import sympy

mean_x_v = sympy.Symbol('mean_x_v');
mean_y_v = sympy.Symbol('mean_y_v');
mean_xx_v = sympy.Symbol('mean_xx_v');
mean_xy_v = sympy.Symbol('mean_xy_v');
mean_yy_v = sympy.Symbol('mean_yy_v');

slope_v = sympy.Symbol('slope_v');
intercept_v = sympy.Symbol('intercept_v');

ss_tot   = mean_yy_v - mean_y_v * mean_y_v;

ss_res = (
     mean_yy_v - 2 * slope_v * mean_xy_v - 2 * intercept_v * mean_y_v + 
    slope_v * slope_v * mean_xx_v + 2 * slope_v * intercept_v * mean_x_v + 
    intercept_v * intercept_v);

ss_tot_m_res = (slope_v * (
            (mean_xy_v - slope_v * mean_xx_v) +
            (mean_xy_v - intercept_v * mean_x_v)
        ) + intercept_v * (mean_y_v - slope_v * mean_x_v - intercept_v) + 
       mean_y_v * (intercept_v - mean_y_v));

diff = sympy.simplify( ss_tot_m_res - (ss_tot - ss_res))
assert diff == sympy.core.numbers.Zero()
```

nvbench/detail/online_linear_regression.cuh
```diff
@@ -0,0 +1,185 @@
...
+    {
+      return std::numeric_limits<nvbench::float64_t>::quiet_NaN();
+    }
+
+    // ss_tot and ss_res scaled by 1/n to avoid overflow
+    const nvbench::float64_t n        = static_cast<nvbench::float64_t>(m_count);
+    const nvbench::float64_t mean_y_v = mean_y();
+    const nvbench::float64_t ss_tot   = (m_sum_y2 / n) - mean_y_v * mean_y_v;
+
+    if (ss_tot == 0)
```
oleksandr-pavlyk:
```suggestion
    if (ss_tot < 1e-9)
```

Changing this to `ss_tot == 0` breaks  entropy_criterion's `test_const`: https://github.com/NVIDIA/nvbench/blob/main/testing/entropy_criterion.cu#L29-L40

jhou-jpg: was worried the 1e-9 tolerance might be too forgiving, given these are gonna be handling math for kernel running 0.00x ms

oleksandr-pavlyk: Added an output inside the `if` branch here to get a sense of the `ss_tot` value coming from the `test_const` on my machine:

```
(nvbench) ~/repos/nvbench/build-review$ ./bin/nvbench.test.entropy_criterion 
Small ss_tot: 6.84775e-33
m_count: 6
m_sum_x: 15
m_sum_y: 2.22045e-16
m_sum_x2: 55
m_sum_xy: 6.66134e-16
m_sum_y2: 4.93038e-32
```

With constant time measurements of 42.0 used by the test, the exact value of the entropy remains zero. `compute_entropy` returns a non-zero value due to round-off errors.  

I think it is safe to use `ss_tot < std::numeric_limits<nvbench::float64_t>::epsilon()`.

nvbench/detail/online_linear_regression.cuh
oleksandr-pavlyk: This change is required to fix build break.

nvbench/detail/online_linear_regression.cuh
oleksandr-pavlyk: This change is just a suggestion, can be deferred to a subsequent PR. Feel free to ignore.

nvbench/detail/online_linear_regression.cuh
oleksandr-pavlyk: The test failure must be fixed. The likely reason for why it round-off error was introduced is due to my suggestion here: https://github.com/NVIDIA/nvbench/pull/286#discussion_r2550679330 

I believe that `ss_tot < std::numeric_limits<nvbench::float64_t>::epsilon()` is safe to use.

PR #289, opened 2025-12-04
Add multi-cuda wheel build

shwina:
This PR introduces "multi-CUDA" wheels builds, exactly similar to the ones we build for CCCL, as described [here](https://github.com/NVIDIA/cccl/pull/5613).

Instead of shipping multiple packages (`pynvbench-cu12`, `pynvbench-cu13`); we ship a single one (`pynvbench`) that contains both CUDA 12 and 13 binaries (bindings). The resulting wheel size is just ~2.2MB, so this is not a problem. At runtime, depending on the available CUDA version, the appropriate binary is loaded.

At install time, we still need the user to specify which CUDA major version to install for - because the version of required dependencies (`nvidia-cuda-cupti-cu12`) depend on the CUDA major version. This specification of the CUDA version is done using "extras":

```
pip install pynvbench[cu12]  # or cu13
```

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -30,10 +30,10 @@ readme = { file = "README.md", content-type = "text/markdown" }
 
 [project.optional-dependencies]
 # CUDA 12.x dependencies
-cu12 = ["nvidia-cuda-cupti-cu12"]
+cu12 = ["cuda-bindings>=12.0.0,<13.0.0", "nvidia-cuda-cupti-cu12"]
 
 # CUDA 13.x dependencies
-cu13 = ["nvidia-cuda-cupti>=13.0"]
+cu13 = ["cuda-bindings>=13.0.0,<14.0.0", "nvidia-cuda-cupti>=13.0"]
 
 # Test dependencies for CUDA 12
 test-cu12 = ["pynvbench[cu12]", "pytest", "cupy-cuda12x", "numba"]


python/pyproject.toml
```diff
@@ -22,22 +22,26 @@ classifiers = [
...
-  # The shared library
-  # libnvidia-ml must be installed system-wide
-  # (Debian package provider: libnvidia-compute)
 ]
 dynamic = ["version"]
 readme = { file = "README.md", content-type = "text/markdown" }
 
 [project.optional-dependencies]
+# CUDA 12.x dependencies
+cu12 = ["cuda-bindings>=12.0.0,<13.0.0", "nvidia-cuda-cupti-cu12"]
```
shwina, line 33: Remove the `cuda-bindings` dependency as it's just used as a proxy for the CUDA major version right now.

python/CMakeLists.txt
```diff
@@ -25,17 +25,23 @@ CPMAddPackage(
 
 CPMAddPackage("gh:pybind/pybind11@3.0.0")
 
+# Determine CUDA major version for directory structure
+string(REGEX MATCH "^([0-9]+)" CUDA_VERSION_MAJOR "${CUDAToolkit_VERSION}")
```
alliepiper: nit: `CUDAToolkit_VERSION_MAJOR` should already be defined as part of CUDAToolkit.

ci/build_multi_cuda_wheel.sh
```diff
@@ -0,0 +1,191 @@
...
+readonly devcontainer_version=25.12
+readonly devcontainer_distro=rockylinux8
+
+if [[ "$(uname -m)" == "aarch64" ]]; then
+  readonly cuda12_image=rapidsai/ci-wheel:${devcontainer_version}-cuda${cuda12_version}-${devcontainer_distro}-py${py_version}-arm64
+  readonly cuda13_image=rapidsai/ci-wheel:${devcontainer_version}-cuda${cuda13_version}-${devcontainer_distro}-py${py_version}-arm64
+else
+  readonly cuda12_image=rapidsai/ci-wheel:${devcontainer_version}-cuda${cuda12_version}-${devcontainer_distro}-py${py_version}
+  readonly cuda13_image=rapidsai/ci-wheel:${devcontainer_version}-cuda${cuda13_version}-${devcontainer_distro}-py${py_version}
+fi
```
oleksandr-pavlyk, line 60: Nit: Perhaps introduce `readonly host_arch_suffix` to set to `-arm64` or leave empty` and then image names could be set unconditionally, and arch names would only need to be changed in one place:

```sh
readonly cuda12_image=rapidsai/ci-wheel:${devcontainer_version}-cuda${cuda12_version}-${devcontainer_distro}-py${py_version}${arch_suffix}

readonly cuda13_image=rapidsai/ci-wheel:${devcontainer_version}-cuda${cuda13_version}-${devcontainer_distro}-py${py_version}${arch_suffix}
```

python/pyproject.toml
```diff
@@ -22,22 +22,26 @@ classifiers = [
...
+# CUDA 13.x dependencies
+cu13 = ["cuda-bindings>=13.0.0,<14.0.0", "nvidia-cuda-cupti>=13.0"]
+
+# Test dependencies for CUDA 12
+test-cu12 = ["pynvbench[cu12]", "pytest", "cupy-cuda12x", "numba"]
+
+# Test dependencies for CUDA 13
+test-cu13 = ["pynvbench[cu13]", "pytest", "cupy-cuda13x", "numba"]
+
+# Generic test dependencies (defaults to CUDA 12)
```
oleksandr-pavlyk, line 44: Shouldn't we default to the latest major version of CTK? Is there a reason not to default to using cu13?

Issue #295, opened 2025-12-10
Allow partial comparison in `nvbench_compare.py`


bernhardmgruber:
When I run the same benchmark, once full and once a subset (to save execution time):
```
./bin/cub.bench.transform.babelstream.base -d 0 --stopping-criterion entropy --json base.json
./bin/cub.bench.transform.babelstream.base -d 0 -a "T{ct}=I16" -a "Elements{io}[pow2]=28" --stopping-criterion entropy --json new.json
```
and then compare the results:
```
PYTHONPATH=./_deps/nvbench-src/scripts python3 ./_deps/nvbench-src/scripts/nvbench_compare.py ./base.json  ./new.json
['./base.json', './base_bif.json']
# Summary

- Total Matches: 0
  - Pass    (diff <= min_noise): 0
  - Unknown (infinite noise):    0
  - Failure (diff > min_noise):  0

```
I get an empty comparison results. I think the script should report the result for each benchmark which is present in both JSON files and only skip the ones present in only one JSON file.

The current workaround is to create a partial `base.json` as well, which is 100% unnecessary.

PR #296, opened 2025-12-10
Allow partial comparison in `nvbench_compare.py`

bernhardmgruber:
Fixes: #295

Can compare the files in #295:
```
['base.json', 'base_bif.json']
# mul

## [0] NVIDIA H200

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |     Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|----------|---------|----------|
|   I16   |      I32      |      2^28      | 253.092 us |       0.42% | 259.197 us |       0.35% | 6.104 us |   2.41% |   SLOW   |
|   I16   |      I64      |      2^28      | 253.069 us |       0.42% | 259.680 us |       0.37% | 6.612 us |   2.61% |   SLOW   |

# add

## [0] NVIDIA H200

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |     Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|----------|---------|----------|
|   I16   |      I32      |      2^28      | 371.395 us |       0.22% | 374.366 us |       0.24% | 2.972 us |   0.80% |   SLOW   |
|   I16   |      I64      |      2^28      | 371.890 us |       0.21% | 374.705 us |       0.25% | 2.815 us |   0.76% |   SLOW   |

# triad

## [0] NVIDIA H200

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |     Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|----------|---------|----------|
|   I16   |      I32      |      2^28      | 373.139 us |       0.21% | 374.479 us |       0.24% | 1.340 us |   0.36% |   SLOW   |
|   I16   |      I64      |      2^28      | 373.672 us |       0.21% | 374.310 us |       0.24% | 0.638 us |   0.17% |   SAME   |

# nstream

## [0] NVIDIA H200

|  T{ct}  |  OffsetT{ct}  |  Elements{io}  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |      Diff |   %Diff |  Status  |
|---------|---------------|----------------|------------|-------------|------------|-------------|-----------|---------|----------|
|   I16   |      I32      |      2^28      | 492.228 us |       0.08% | 488.943 us |       0.15% | -3.285 us |  -0.67% |   FAST   |
|   I16   |      I64      |      2^28      | 492.258 us |       0.13% | 489.330 us |       0.16% | -2.928 us |  -0.59% |   FAST   |

# Summary

- Total Matches: 8
  - Pass    (diff <= min_noise): 1
  - Unknown (infinite noise):    0
  - Failure (diff > min_noise):  7
```

### DIFF
--- a/python/scripts/nvbench_compare.py
+++ b/python/scripts/nvbench_compare.py
@@ -32,7 +32,7 @@ pass_count = 0
 
 def find_matching_bench(needle, haystack):
     for hay in haystack:
-        if hay["name"] == needle["name"] and hay["axes"] == needle["axes"]:
+        if hay["name"] == needle["name"]:
             return hay
     return None
 


python/scripts/nvbench_compare.py
```diff
@@ -27,7 +27,7 @@ def version_tuple(v):
 
 def find_matching_bench(needle, haystack):
     for hay in haystack:
-        if hay["name"] == needle["name"] and hay["axes"] == needle["axes"]:
+        if hay["name"] == needle["name"]:
```
oleksandr-pavlyk, line 35: Should we check that intersection of `hay["axes"]` and `need["axes"]` is not empty?

oleksandr-pavlyk: I had something as follows in mind:

```python
def axes_do_intersect(hay, needle):
    if (len(hay) != len(needle)):
        return False
    for ha, na in zip(hay, needle):
        if any(ha[k] != na[k] for k in ["name", "type", "flags"]):
            return False
        hv, nv = ha["values"], na["values"]
        if not any(v in hv for v in nv):
            return False
    return True


def find_matching_bench(needle, haystack):
    for hay in haystack:
        if hay["name"] == needle["name"] and axes_do_intersect(hay["axes"], needle["axes"]):
            return hay
    return None
```

oleksandr-pavlyk: @bernhardmgruber Ping, do you object to this change? I think it is necessary to check that the sets of axes values for hay and needle do intersect, otherwise the script output is not going to be meaningful.

oleksandr-pavlyk (approved):
Approving as is. I will open another PR to add `axes_do_intersect` and will provide an example of why this is necessary based on workflow from #295

bernhardmgruber:
Sorry, I haven't found time to work on this :S But I do repeatedly need this feature. Another example: we benchmark a full baseline for a CUB algorithm, then notice only `I8` workloads regress, so we want to keep benchmarking and comparing only `I8` workloads. here, partial comparisons would again we very useful.

> Approving as is. I will open another PR to add `axes_do_intersect` and will provide an example of why this is necessary based on workflow from #295

I don't mind if we alter this PR or feature, I just want to be able to get a meaningful comparison where one file just has less runs than the other.

python/scripts/nvbench_compare.py
bernhardmgruber: No objection. Can you implement this for me please and see whether it still produces a meaningful diff for the files attached to #295? Thx!

oleksandr-pavlyk: I tried processing files from #295 with these changes, and they produce the same output as with your changes. Where suggested change would make a difference is when `data1.json` is collected with `-a "T{ct}=I8"` and `data2.json` is collected with `-a "T{ct}=I32"`.

PR #308, opened 2026-01-30
Add 89-real to list of architectures built for cuda-bench

NaderAlAwar:
When building python wheels, we mistakenly left out compiling for `sm_89`. This PR adds this to the compilation script.


NaderAlAwar:
@oleksandr-pavlyk good point, I'll edit the original description

PR #311, opened 2026-02-02
Remove cupti from cuda-bench dependencies

NaderAlAwar:
We currently include an option in NVBench that enables collecting cupti metrics. Following some internal discussion, we decided to remove this option as it is meant for profiling and is orthogonal to NVBench's goal which is to benchmark kernels.


NaderAlAwar:
pre-commit.ci autofix

PR #314, opened 2026-02-05
Add a script to plot benchmark results

bernhardmgruber:
This script is 100% vibe-coded. I only looked at the result, since I don't understand Python. Here is a generated summary of what this PR proposes:
```
Add nvbench_plot.py to plot horizontal bar charts from NVBench JSON results using %SOL (global BW utilization).
Support filtering by axis values (-a with optional [pow2]) and benchmark name (-b).
Improve plot presentation with per-benchmark colors, tighter vertical spacing, and sensible default title.
```
Examples filtering by axis and benchmark name:
```
python ./python/scripts/nvbench_plot_bwutil.py -b nstream -b triad -a Elements{io}[pow2]=28 .
/babelstream_ublkcp_B200.json
```
image

Examples with two files, filtering by axis, dark mode:
```
python ./python/scripts/nvbench_plot_bwutil.py -a Elements{io}[pow2]=28 -a T{ct}=F32 ./babels
tream_ublkcp_B200.json ./pytorch_ublkcpy_B200.json --dark
```
image

Here are the input files for the plots:
[pytorch_ublkcpy_B200.json](https://github.com/user-attachments/files/25091855/pytorch_ublkcpy_B200.json)
[babelstream_ublkcp_B200.json](https://github.com/user-attachments/files/25091854/babelstream_ublkcp_B200.json)

Here are the prompts that I used:

1. “Please study the scripts in the directory `python/scripts`. Then look at the benchmark result `babelstream_ublkcp_B200.json`. Create a new script `nvbench_plot.py` in the above directory that can read the benchmark result and plot a horizontal bar chart, where each benchmark state is plotted as a separate box. The box length is taken from `nv/cold/bw/global/utilization`”
2. “If I run the script like `python ./python/scripts/nvbench_plot.py ./babelstream_ublkcp_B200.json` it errors, please fix the error”
3. “Let's add a command line parameter `-a` where I can specify an axis name and value like `-a T{ct}=I8`. When specified only benchmark states are shown where the `axis_value` node in the JSON has an axis named `T{ct}` with a `value` of `I8`”
4. “Please add special handling to the `-a` argument handling where when I enter an like `-a Elements{io}[pow2]=28` the `[pow2]` is parsed before the `=` and it converts the value after the `=` (`28` in this case) to 2 to the power of that value (`2^28` in this case)”
5. “Let's add another command line parameter `-b` where i can select the benchmark name”
6. “Make the bar color #76b900”
7. “Please do not plot the file name next to the bar charts”
8. “Please include the device name in the bar labels”
9. “Let's remove the device name again”
10. “Please remove the `Device=N` part from the bar label”
11. “Can you use different colors for the bars per benchmark.”
12. “Add a plot tile "%SOL Bandwidth". If the device of all benchmarks is the same, put the device name in the benchmark title like "%SOL Bandwidth - B200"”
13. “I get the warning

The get_cmap function was deprecated in Matplotlib 3.7 and will be removed in 3.11. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap()`` or ``pyplot.get_cmap()`` instead.

Please fix it”
14. “Please reduce the empty space above the first bar and below the last bar”
15. “Please write a pull request description about what this new script can do”
16. "Add --dark mode with black background and white text."
17. "Ensure title is visible in dark mode."
18. "Remove the x‑axis description label."
19. "If -a is used, remove those axis tokens from labels and add them to the title."
20. “Please read through the nvbench code here and figure out whether it supports -a paramters where we can also specify multiple values”
21. “Please implement this”
22. “If multiple values are specified for an axis, retain the axis value in the box label”

### DIFF
--- a/python/scripts/nvbench_plot_bwutil.py
+++ b/python/scripts/nvbench_plot_bwutil.py
@@ -12,7 +12,6 @@ try:
 except ImportError:
     from scripts.nvbench_json import reader
 
-
 UTILIZATION_TAG = "nv/cold/bw/global/utilization"
 
 
@@ -55,12 +54,11 @@ def parse_files():
     for file_or_dir in files_or_dirs:
         if os.path.isdir(file_or_dir):
             for f in os.listdir(file_or_dir):
-                if os.path.splitext(f)[1] != ".json":
-                    continue
                 filename = os.path.join(file_or_dir, f)
                 if os.path.isfile(filename) and os.path.getsize(filename) > 0:
                     filenames.append(filename)
         else:
+            assert os.path.isfile(file_or_dir)
             filenames.append(file_or_dir)
 
     filenames.sort()
@@ -107,7 +105,9 @@ def parse_axis_filters(axis_args):
         if value.startswith("[") and value.endswith("]"):
             inner = value[1:-1].strip()
             if inner:
-                values = [item.strip() for item in inner.split(",") if item.strip()]
+                values = [
+                    stripped for item in inner.split(",") if (stripped := item.strip())
+                ]
             else:
                 values = []
         else:
@@ -189,7 +189,9 @@ def strip_axis_filters_from_state_name(state_name, axis_filters):
     return " ".join(tokens)
 
 
-def collect_entries(filename, axis_filters, benchmark_filters):
+def collect_entries(
+    filename: str, axis_filters: list[dict], benchmark_filters: list[str]
+) -> tuple[list[tuple[str, float, str]], set[str]]:
     json_root = reader.read_file(filename)
     entries = []
     device_names = set()
@@ -225,14 +227,8 @@ def plot_entries(entries, title=None, output=None, dark=False):
         print("No utilization data found.")
         return 1
 
-    labels = [entry[0] for entry in entries]
-    values = [entry[1] for entry in entries]
-    bench_names = [entry[2] for entry in entries]
-
-    unique_benches = []
-    for bench in bench_names:
-        if bench not in unique_benches:
-            unique_benches.append(bench)
+    labels, values, bench_names = map(list, zip(*entries))
+    unique_benches = list(set(bench_names))
 
     cmap = plt.get_cmap("tab20", max(len(unique_benches), 1))
     bench_colors = {bench: cmap(index) for index, bench in enumerate(unique_benches)}


python/scripts/nvbench_plot.py
```diff
@@ -0,0 +1,299 @@
...
+        help="Filter by benchmark name (can repeat)",
+    )
+    args, files_or_dirs = parser.parse_known_args()
+
+    filenames = []
+    for file_or_dir in files_or_dirs:
+        if os.path.isdir(file_or_dir):
+            for f in os.listdir(file_or_dir):
+                if os.path.splitext(f)[1] != ".json":
+                    continue
```
oleksandr-pavlyk: I'd rather not filter out files based on their extensions. Instead we should print meaningful error message when JSON parsing throws an error.

I often collect data using `--jsonbin run1` which creates JSON file `run1` without any extension at all.

python/scripts/nvbench_plot_bwutil.py
```diff
@@ -0,0 +1,299 @@
...
+    for file_or_dir in files_or_dirs:
+        if os.path.isdir(file_or_dir):
+            for f in os.listdir(file_or_dir):
+                if os.path.splitext(f)[1] != ".json":
+                    continue
+                filename = os.path.join(file_or_dir, f)
+                if os.path.isfile(filename) and os.path.getsize(filename) > 0:
+                    filenames.append(filename)
+        else:
+            filenames.append(file_or_dir)
```
oleksandr-pavlyk, line 62: Add `assert os.path.isfile(file_or_dir)`

python/scripts/nvbench_plot.py
```diff
@@ -0,0 +1,299 @@
...
+    )
+    tokens = [
+        token
+        for token in tokens
+        if not any(token.startswith(prefix) for prefix in filter_prefixes)
+    ]
+    return " ".join(tokens)
+
+
+def collect_entries(filename, axis_filters, benchmark_filters):
```
oleksandr-pavlyk: Suggestion to add type annotations, both here and throughout. This self-documents the code and allows for more efficient use of Python linters.

```suggestion
def collect_entries(filename : str, axis_filters : list[dict], benchmark_filters: list[str]) -> tuple[list[tuple[str, float, str]], set[str]]:
```

python/scripts/nvbench_plot.py
```diff
@@ -0,0 +1,299 @@
...
+        value = value.strip()
+        if not name or not value:
+            raise ValueError("Axis filter must be NAME=VALUE: {}".format(axis_arg))
+
+        values = []
+        display_values = []
+        if value.startswith("[") and value.endswith("]"):
+            inner = value[1:-1].strip()
+            if inner:
+                values = [item.strip() for item in inner.split(",") if item.strip()]
```
oleksandr-pavlyk:
```suggestion
                values =[stripped for item in inner.split(",") if (stripped := item.strip())]
```

python/scripts/nvbench_plot.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def plot_entries(entries, title=None, output=None, dark=False):
+    if not entries:
+        print("No utilization data found.")
+        return 1
+
+    labels = [entry[0] for entry in entries]
+    values = [entry[1] for entry in entries]
+    bench_names = [entry[2] for entry in entries]
```
oleksandr-pavlyk:
```suggestion
    # unpack entries
    labels, values, bench_names = map(list, zip(*entries))
```

python/scripts/nvbench_plot.py
```diff
@@ -0,0 +1,299 @@
...
+
+    unique_benches = []
+    for bench in bench_names:
+        if bench not in unique_benches:
+            unique_benches.append(bench)
```
oleksandr-pavlyk:
```suggestion
    unique_benches = list(set(bench_names))
```

python/scripts/nvbench_plot.py
bernhardmgruber: I removed the check.

PR #332, opened 2026-04-01
Add `--no-color` flag to nvbench_compare.py which can be use for github issues and PRs

NaderAlAwar:
Update: renamed the flag to `--no-color`

The status column in the markdown table generated by `nvbench_compare.py` currently color codes the output (e.g., green for FAST). This does not get formatted well in github issues and PRs. For example, in a recent CCCL benchmakr CI run (https://github.com/NVIDIA/cccl/actions/runs/23866054665, scroll down and expand either the CUB or Python comparison reports), the output appears as �[34mSAME�[39m

This PR adds a flag which color codes the status column using an emoji:

|   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |      Diff |   %Diff |  Status  |
|------------|-------------|------------|-------------|-----------|---------|----------|
|   1.003 ms |       0.06% |   1.003 ms |       0.05% | -0.071 us |  -0.01% | 🔵 SAME  |
|   1.003 ms |       0.03% |   1.003 ms |       0.03% |  0.049 us |   0.00% | 🔵 SAME  |

### DIFF
--- a/python/scripts/nvbench_compare.py
+++ b/python/scripts/nvbench_compare.py
@@ -285,7 +285,7 @@ def compare_benches(
     dark,
     axis_filters,
     benchmark_filters,
-    markdown=False,
+    no_color,
 ):
     if plot_along:
         import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


python/scripts/nvbench_compare.py
```diff
@@ -594,6 +607,13 @@ def main():
         action="store_true",
         help="Use dark theme (black background, white text)",
     )
+    parser.add_argument(
+        "--markdown",
+        dest="markdown",
+        default=False,
+        action="store_true",
+        help="Use emoji instead of ANSI color codes (useful for GitHub issues/PRs)",
+    )
```
bernhardmgruber, line 615: Suggestion: Since the original output is also markdown, I don't think `--markdown` is the right name for this option. Can we name it something like `--nocolor`?

NaderAlAwar: Fair point, I actually named it nocolor at first. I can switch back to that

bernhardmgruber (approved):

alliepiper:
This has been so long overdue, thanks for adding it 🙏

PR #372, opened 2026-05-20
Add frequency measurement to the stopping criterion

mfranzrebsal:
This MR makes every iteration's measured average frequency available to the stopping criterion, in case it wants to use it to make decisions. The virtual function is implemented as a no-op so that classes that do not have it implemented do not break. If you think it would be a better idea to implement it some other way, please let me know!

On another note: I have noticed that you are using FP32 for the frequency, which is always going to be an integer since it is in Hz, and for values in the GHz range, FP32 has a granularity of around 128. I don't think this is especially problematic given that when the frequency is in the order of GHz, a discrepancy of 128 Hz is insignificant, but wanted to bring it to your attention just in case.

### DIFF
--- a/examples/frequency_criterion.cu
+++ b/examples/frequency_criterion.cu
@@ -1,0 +1,120 @@
+/*
+ *  Copyright 2026 NVIDIA Corporation
+ *
+ *  Licensed under the Apache License, Version 2.0 with the LLVM exception
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.
+ *
+ *  You may obtain a copy of the License at
+ *
+ *      http://llvm.org/foundation/relicensing/LICENSE.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include <nvbench/nvbench.cuh>
+
+// Grab some testing kernels from NVBench:
+#include <nvbench/test_kernels.cuh>
+
+// Thrust vectors simplify memory management:
+#include <thrust/device_vector.h>
+
+#include <stdexcept>
+
+// This example shows how to write a stopping criterion that *requires* the GPU
+// clock frequency that NVBench observes for every cold-measurement sample.
+//
+// In addition to `do_add_measurement()`, a criterion may override
+// `do_add_frequency()` to receive the SM clock rate (in Hz) measured during the
+// sample. NVBench calls `add_frequency()` immediately before `add_measurement()`
+// for the same sample -- but only when it can measure the clock. It is NOT
+// called while profiling (the `--profile` option) or for CPU-only benchmarks
+// (`nvbench::exec_tag::cpu_only` / `no_gpu`).
+//
+// Like the `fixed` criterion in `custom_criterion.cu`, this one simply runs for
+// a fixed number of samples. The difference is that it also collects the
+// per-sample frequency and throws if a sample arrives without one. The thrown
+// exception is caught per-benchmark by NVBench and reported as a failure, so
+// running this benchmark with `--profile` produces a clear error instead of
+// silently ignoring the missing frequency.
+
+// Inherit from the stopping_criterion_base class:
+class frequency_criterion final : public nvbench::stopping_criterion_base
+{
+  nvbench::int64_t m_num_samples{};
+  bool m_has_frequency{false};
+
+public:
+  frequency_criterion()
+      : nvbench::stopping_criterion_base{"frequency", {{"max-samples", nvbench::int64_t{42}}}}
+  {}
+
+protected:
+  // Setup the criterion in the `do_initialize()` method:
+  virtual void do_initialize() override
+  {
+    m_num_samples   = 0;
+    m_has_frequency = false;
+  }
+
+  // Collect the GPU clock frequency for the current sample. NVBench calls this
+  // before `do_add_measurement()` whenever a frequency is available:
+  virtual void do_add_frequency(nvbench::float32_t /* frequency_hz */) override
+  {
+    m_has_frequency = true;
+  }
+
+  // Process new measurements in the `do_add_measurement()` method:
+  virtual void do_add_measurement(nvbench::float64_t /* measurement */) override
+  {
+    // This criterion requires a frequency for every sample. NVBench calls
+    // `do_add_frequency()` before `do_add_measurement()` when one is available,
+    // so a missing frequency here means none was provided for this sample:
+    if (!m_has_frequency)
+    {
+      throw std::runtime_error(
+        "frequency_criterion requires a GPU clock frequency for every sample, but none was "
+        "provided. NVBench does not measure the clock frequency when profiling (--profile) or for "
+        "CPU-only benchmarks (nvbench::exec_tag::cpu_only / no_gpu).");
+    }
+
+    m_has_frequency = false; // consume it; the next sample must provide its own
+    m_num_samples++;
+  }
+
+  // Check if the stopping criterion is met in the `do_is_finished()` method:
+  virtual bool do_is_finished() override
+  {
+    return m_num_samples >= m_params.get_int64("max-samples");
+  }
+};
+
+// Register the criterion with NVBench:
+NVBENCH_REGISTER_CRITERION(frequency_criterion);
+
+void throughput_bench(nvbench::state &state)
+{
+  // Allocate input data:
+  const std::size_t num_values = 64 * 1024 * 1024 / sizeof(nvbench::int32_t);
+  thrust::device_vector<nvbench::int32_t> input(num_values);
+  thrust::device_vector<nvbench::int32_t> output(num_values);
+
+  // Provide throughput information:
+  state.add_element_count(num_values, "NumElements");
+  state.add_global_memory_reads<nvbench::int32_t>(num_values, "DataSize");
+  state.add_global_memory_writes<nvbench::int32_t>(num_values);
+
+  state.exec([&input, &output, num_values](nvbench::launch &launch) {
+    (void)num_values; // clang thinks this is unused...
+    nvbench::copy_kernel<<<256, 256, 0, launch.get_stream()>>>(
+      thrust::raw_pointer_cast(input.data()),
+      thrust::raw_pointer_cast(output.data()),
+      num_values);
+  });
+}
+NVBENCH_BENCH(throughput_bench).set_stopping_criterion("frequency");
--- a/nvbench/detail/measure_cold.cu
+++ b/nvbench/detail/measure_cold.cu
@@ -147,6 +147,8 @@ void measure_cold_base::record_measurements()
 
     m_sm_clock_rates.push_back(current_clock_rate);
     m_sm_clock_rate_accumulator += current_clock_rate;
+
+    m_stopping_criterion.add_frequency(current_clock_rate);
   }
 
   // Update and record timers and counters:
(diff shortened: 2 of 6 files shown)


oleksandr-pavlyk:
@coderabbitai full review

nvbench/detail/measure_cold.cu
```diff
@@ -101,9 +101,9 @@ void measure_cold_base::run_trials_prologue() { m_walltime_timer.start(); }
 
 void measure_cold_base::record_measurements()
 {
+  const auto current_clock_rate = m_gpu_frequency.get_clock_frequency();
```
oleksandr-pavlyk: For the `m_gpu_frequency.get_clock_frequency()` return value to be meaningful, calls to `m_gpu_frequency.start()` and `m_gpu_frequency.stop()` must be made to populate internal time-point data members.  However, this is only done when `m_check_throttling` is true.  (See nvbench/detail/measure_cold_launch_timer_core.cuh:119 and nvbench/detail/measure_cold_launch_timer_core.cuh:146).

For instance, when `--profile` is used,  `m_check_throttling` value would be `false`, the frequency timestamp recording would be skipped and the return value of `get_clock_frequency()` would be based on the content of uninitialized memory. 

This unspecified value would be passed to the stopping criterion object.

nvbench/detail/measure_cold.cu
```diff
@@ -165,6 +165,7 @@ void measure_cold_base::record_measurements()
 
   ++m_total_samples;
 
+  m_stopping_criterion.add_frequency(current_clock_rate);
   m_stopping_criterion.add_measurement(cur_cuda_time);
```
oleksandr-pavlyk: `stopping_criterion_base::add_frequency()` is added as a separate callback.

The interface does not document the ordering/pairing contract.

The cold measurement calls `add_frequency()` before `add_measurement()` only for accepted samples, and CPU-only measurement does not call it at all.

Ideally, all stopping criterion classes should work with all measurement classes.

nvbench/detail/measure_cold.cu
oleksandr-pavlyk: We could modify `measure_cold_launch_timer_core.cuh` so that the start and stop methods be called unconditionally.

mfranzrebsal: Changed in the last commit.

nvbench/detail/measure_cold.cu
mfranzrebsal: I have added more documentation in the last commit. What exactly do you mean by "all stopping criterion classes should work with all measurement classes"? Is that not the case currently?

nvbench/detail/measure_cold.cu
mfranzrebsal: One thought on this: with this change, we always measure frequency, and when calling `measure_cold_launch_timer_core::stop` when profiling, `gpu_frequency_stop` will be called before `profiler_stop`, meaning that `get_timestamps_kernel` will be collected. Shall I change the order of that?

oleksandr-pavlyk: The profiler should start **before kernel is launched** and stop **after host synchronizes** with device.

The call to `sync_stream()` in `stop()` method serves to ensure completion on both the kernel of interest and of utility `device_timestamps_kernel`.  Moving `profiler_stop()` before `gpu_frequency_stop()` would require us to use two synchronization points:

```
void stop()
{
   cuda_timer_stop();
   unblock_stream();
   sync_stream();
   profiler_stop();
   gpu_frequency_stop();
   sync_stream();
   cpu_timer_stop();
}
```

This would slow down all benchmarks and is undesirable.

So I think moving `gpu_frequency_stop()` from `m_check_trottling` guard was a mistake. The guard ensured that `device_timestamps_kernel` is not launched when `--profiler` is used.

This brings the issue back. The stopping criterion likely needs to accept `std::optional<nvbench::float32_t>` instead of `nvbench::float32_t` directly, and the `stopping_criterion` class should make sure that `std::nullopt` is passed when gpu_frequency data were not yet initialized. 

Perhaps `gpu_frequency` struct needs an additional boolean member variable to track whether it has been started and stopped.

mfranzrebsal: Okay, can implement that. What if:
```
void stop()
{
   cuda_timer_stop();
   profiler_stop(); <- Sync inside the if statement
   gpu_frequency_stop();
   unblock_stream();
   sync_stream();
   cpu_timer_stop();
}
```

This would keep the ordering and sync behavior when not running in profile mode. The possible issue I'm seeing with the `optional` version is that my stopping criterion would have to behave differently when frequencies are being collected vs when not. Currently frequencies are only not collected when `run_once = true` i.e. when profiling, so that's not a problem, but it could become one in the future. On the other hand, I don't see issues like that arising from always collecting frequencies. WDYT?

oleksandr-pavlyk: > The possible issue I'm seeing with the optional version is that my stopping criterion would have to behave differently when frequencies are being collected vs when not.

I am afraid that is necessary irrespective of how we change `stop()` method. Currently, a user can compile benchmark with arbitrary execution tag, including `exec_tag::cpu_only`, The `measure_cpu_only` does not supply stopping criterion with frequency data. 

To make this clean either any stopping criterion should handle absence of GPU frequency signal, or such a code should not compile. This is what I meant by "all stopping criterion classes should work with all measurement classes".

mfranzrebsal: I've added a change that I think is simpler than using `std::optional` and should have similar characteristics. Let me know what you think!

nvbench/detail/measure_cold.cu
oleksandr-pavlyk: > What exactly do you mean by "all stopping criterion classes should work with all measurement classes"?

Author implementing a stopping criterion that takes advantage of frequency information:
   1. Ensures that criterion handles situation where frequency data is not available. That would occur if a benchmark uses CPU-only timer, e.g. benchmark uses `state.exec(nvbench::exec_tag::cpu_only, ...);`.   
   2. Handling GPU frequency data:
        - Author must be assured that number of frequency measurements processed is either 0 or is equal to the number of sample measurements processed.
        - Author must be informed of any consistency conditions expected by NVBench of the criterion. For example, we might expect that absent GPU data and a stream of sample measurements, criterion should behave as if GPU frequency data is available but contain a constant value. 
        - Author must be instructed of a mechanism for stopping criterion to enforce that frequency data is provided. For example, it might throw an exception if frequency and sample counts disagree. We need to make sure that an NVBench-instrumented benchmark handles such situation well.

mfranzrebsal: Re 1: I agree, but that is fully the users responsibility, correct?
Re 2:
- How I see it is that currently only two things can happen: either we never add any frequency measurements, or we add as many as we do timing measurements. So if the number of frequency measurements is larger than 0 but not the same as the number of timing measurements, that would be a bug.
- What do you mean by "we might expect that absent GPU data and a stream of sample measurements, criterion should behave as if GPU frequency data is available but contain a constant value."?
- From a user's perspective, creating a stopping criterion that only works when frequency data is provided is brittle IMO, and should not be encouraged or supported.

Of course, I trust your judgement on this more. What changes do you think are still necessary?

oleksandr-pavlyk: The issue is that personas writing the benchmark and running benchmarks may be different. The choice of measuring policy to to use is decided by the former, while the stopping criterion can be set at the runtime. 

That is unless the criterion is not set using `nvbench::benchmark::set_criterion`, and I am yet to verify whether the CLI options would override that choice. But regardless, we need to check that throwing an exception in user-defined frequency collecting stopping criterion actually terminates the collection. I suspect throwing an exception would not be enough. The stopping criterion must warn and flag measurements as finished.

I would prefer to have concrete C++ examples on top of this PR to have answers to these questions, and have a prototypical example of frequency-collecting stopping criterion (does not have to do anything useful either, but illustrate workings of how to it handles being used with cpu-only measurements). I will try to write some up.

mfranzrebsal: I have added an example and test to ensure that when throwing from `add_frequency` or `add_measurement`, the benchmark aborts with an appropriate error message:
```
$./build/bin/nvbench.example.cpp20.frequency_criterion --profile
# Devices

## [0] `NVIDIA RTX 6000 Ada Generation`
* SM Version: 890 (PTX Version: 890)
* Number of SMs: 142
* SM Default Clock Rate: 2535 MHz
* Global Memory: 47671 MiB Free / 48537 MiB Total
* Global Memory Bus Peak: 864 GB/sec (384-bit DDR @9001MHz)
* Max Shared Memory: 100 KiB/SM, 48 KiB/Block
* L2 Cache Size: 98304 KiB
* Maximum Active Blocks: 24/SM
* Maximum Active Threads: 1536/SM, 1024/Block
* Available Registers: 65536/SM, 65536/Block
* ECC Enabled: No

# Log

Run:  [1/1] throughput_bench [Device=0]
Fail: Unexpected error: frequency_criterion requires a GPU clock frequency for every sample, but none was provided. NVBench does not measure the clock frequency when profiling (--profile) or for CPU-only benchmarks (nvbench::exec_tag::cpu_only / no_gpu).

# Benchmark Results

## throughput_bench

### [0] NVIDIA RTX 6000 Ada Generation

No data -- check log.
```

testing/criterion_exception.cu
```diff
@@ -0,0 +1,181 @@
...
+
+#include <stdexcept>
+#include <string>
+
+#include "test_asserts.cuh"
+
+// Verifies that an exception thrown by a stopping criterion aborts the
+// benchmark (the state is marked as failed) instead of being silently swallowed.
+
+namespace
```
fbusato: I would remove it if not strictly necessary

nvbench/stopping_criterion.cuh
```diff
@@ -109,6 +109,14 @@ public:
     this->do_initialize();
   }
 
+  /**
+   * Provide the GPU clock frequency (Hz) observed for the current sample. It is not called when
+   * doing CPU-only benchmarking (i.e.: using `nvbench::exec_tag::cpu_only` or
+   * `nvbench::exec_tag::no_gpu`) or profiling (`--profile` option). When called, it is done before
+   * calling `add_measurement` for the same sample.
+   */
+  void add_frequency(nvbench::float32_t frequency_hz) { this->do_add_frequency(frequency_hz); }
```
fbusato, line 119: I think the documentation could be improved as @oleksandr-pavlyk requested, specifying that number of frequency measurements is either 0 or equal to the number of sample measurements

fbusato (approved):
looks good. I left a couple of minor comments

testing/criterion_exception.cu
mfranzrebsal: Done!

nvbench/stopping_criterion.cuh
mfranzrebsal: Done!

oleksandr-pavlyk:
@coderabbitai full review

examples/frequency_criterion.cu
```diff
@@ -0,0 +1,120 @@
...
+  const std::size_t num_values = 64 * 1024 * 1024 / sizeof(nvbench::int32_t);
+  thrust::device_vector<nvbench::int32_t> input(num_values);
+  thrust::device_vector<nvbench::int32_t> output(num_values);
+
+  // Provide throughput information:
+  state.add_element_count(num_values, "NumElements");
+  state.add_global_memory_reads<nvbench::int32_t>(num_values, "DataSize");
+  state.add_global_memory_writes<nvbench::int32_t>(num_values);
+
+  state.exec([&input, &output, num_values](nvbench::launch &launch) {
```
oleksandr-pavlyk, line 112:
```suggestion
  state.exec(nvbench::exec_tag::no_batch, [&input, &output, num_values](nvbench::launch &launch) {
```

Frequency stopping criterion is not going to be invoked in batched mode. This is why `custom_criterion.cu` also uses this tag. If this tag is not used, hot pass will spend 0.5 second on every benchmark.

PR #418, opened 2026-07-18
Fix cuda_timer move ownership

fallintoplace:
## Summary

- transfer `cuda_timer` event ownership in its move operations
- clean up existing destination events during move assignment
- cover move construction and move assignment in the timer test

## Why

`cuda_timer` owns two raw `cudaEvent_t` handles. Its default move operations copy those handles without clearing the source, so destroying the moved-to and moved-from timers destroys the same events twice. Move assignment also overwrites the destination's existing events without destroying them.

The custom moves preserve the timer's original move-only API while ensuring each event has exactly one owner.

## Validation

- `pre-commit run --files nvbench/cuda_timer.cuh testing/cuda_timer.cu`
- `git diff --check`

CUDA tests were not run locally because the development machine does not have a CUDA compiler.

### DIFF
--- a/testing/cuda_timer.cu
+++ b/testing/cuda_timer.cu
@@ -23,7 +23,27 @@
 
 #include <fmt/format.h>
 
+#include <type_traits>
+#include <utility>
+
 #include "test_asserts.cuh"
+
+static_assert(std::is_nothrow_move_constructible_v<nvbench::cuda_timer>);
+static_assert(std::is_nothrow_move_assignable_v<nvbench::cuda_timer>);
+
+void test_move()
+{
+  nvbench::cuda_timer source;
+  nvbench::cuda_timer moved{std::move(source)};
+  nvbench::cuda_timer assigned;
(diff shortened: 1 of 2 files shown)


testing/cuda_timer.cu
```diff
@@ -52,4 +72,8 @@ void test_basic()
   test_basic(stream1, stream2, false);
 }
 
-int main() { test_basic(); }
+int main()
+{
+  test_move();
+  test_basic();
```
oleksandr-pavlyk: Nit: 

```suggestion
  test_basic();
  test_move();
```

oleksandr-pavlyk:
I built NVBench with CUDA 13.3 and GCC 13.3, and ran the test suite. All passed.

oleksandr-pavlyk:
@coderabbitai full review

oleksandr-pavlyk (approved):
Thank you for your contribution, @fallintoplace .

PR #434, opened 2026-07-23
Fix GCC 14 maybe-uninitialized warning

PointKernel:
Fixes a GCC 14 `-Wmaybe-uninitialized` error found in [cuDF CI](https://github.com/rapidsai/cudf/actions/runs/30030700413/job/89286977364) by replacing the temporary empty optional with a named value.


nvbench/runner.cuh
```diff
@@ -87,7 +87,8 @@ struct runner : public runner_base
   {
     if (m_benchmark.m_devices.empty())
     {
-      this->run_device(std::nullopt, skip_remaining);
+      const std::optional<nvbench::device_info> no_device;
+      this->run_device(no_device, skip_remaining);
```
oleksandr-pavlyk, line 91: I think similar change needs to be done in the body of the loop in the else clause. The `device` variable type is `nvbench::device_info`. It is implicitly converted to `std::optional<nvbench::device_info>` temporary and can trigger the same compilation warning.

```diff
-       this->run_device(device, skip_remaining);
+       const std::optional<nvbench::device_info> selected_device{device};
+       this->run_device(selected_device, skip_remaining);
```

oleksandr-pavlyk (approved):

PR #447, opened 2026-08-04
Fix duplicate fmt ALIAS targets when fmt is already provided

### LINKED ISSUE #279: Provided ways to avoid install dependencies
Hi, is it possible to make the dependency fmt to not be GLOBAL TARGETS
since someone may also find_package(fmt),

Since the default nvbench included "GLOBAL TARGETS" fmt, which cause conflicts.

igorpeshansky:
When an enclosing project already provides `fmt::fmt` — imported from an install tree, or built from source via a hard-coded `add_subdirectory` — configuring NVBench fails (see NVIDIA/cudf-spark-jni#4932, where `fmt` comes via `libcudf`):

```
CMake Error at .../_deps/fmt-src/CMakeLists.txt:300 (add_library):
  add_library cannot create ALIAS target "fmt::fmt" because another target
  with the same name already exists.
```

`rapids_cpm_find` used `GLOBAL_TARGETS` to gate a package presence check, so when #281 dropped `GLOBAL_TARGETS fmt::fmt fmt::fmt-header-only` to fix #279, it also disabled this deduplication. Since `fmt::fmt` is imported by `find_package` outside of CPM, the already-added check in CPM didn't fire either.

This PR adds the guard @robertmaynard suggested in https://github.com/NVIDIA/nvbench/issues/279#issuecomment-5058713721 — skip `rapids_cpm_find` when `fmt::fmt` already exists and treat `fmt` as external. An existing `fmt` is recorded via `rapids_export_package` when export sets are enabled, so consumers could still use `find_dependency(fmt)`.

### Testing

Verified with a minimal project that imports `fmt` from an install tree and then adds NVBench as a subproject:

- stock `main`: reproduces both ALIAS errors
- PR branch: reports `NVBench: using externally provided fmt 11.0.2` and builds `libnvbench.so`
- PR branch, `BUILD_SHARED_LIBS=OFF` and `NVBench_ENABLE_INSTALL_RULES=ON`: configures clean, and the generated build/install dependency files both have `find_dependency(fmt)`, with build using the correct `fmt_DIR`
- standalone NVBench with no pre-existing fmt: unchanged, still adds `fmt` 12.2.0 via CPM

### DIFF
--- a/cmake/NVBenchDependencies.cmake
+++ b/cmake/NVBenchDependencies.cmake
@@ -1,29 +1,42 @@
 ################################################################################
 # fmtlib/fmt
+set(fmt_export_set)
 set(export_set_details)
 set(install_fmt OFF)
 if(NOT BUILD_SHARED_LIBS AND NVBench_ENABLE_INSTALL_RULES)
-  set(export_set_details BUILD_EXPORT_SET nvbench-targets
-                         INSTALL_EXPORT_SET nvbench-targets)
+  set(fmt_export_set nvbench-targets)
+  set(export_set_details BUILD_EXPORT_SET ${fmt_export_set}
+                         INSTALL_EXPORT_SET ${fmt_export_set})
   set(install_fmt ON)
 endif()
 
-rapids_cpm_find(fmt 12.2.0 ${export_set_details}
-  CPM_ARGS
-    GIT_REPOSITORY "https://github.com/fmtlib/fmt.git"
-    GIT_TAG "12.2.0"
-    OPTIONS
-      # Force static to keep fmt internal.
-      "BUILD_SHARED_LIBS OFF"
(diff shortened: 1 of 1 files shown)


robertmaynard (approved):

bdice (approved):

oleksandr-pavlyk:
@igorpeshansky I added tests to cover this breakage in #448. I verified that tests do not pass in main and pass after changes here

oleksandr-pavlyk (approved):

igorpeshansky:
Thanks, @oleksandr-pavlyk! I deliberately avoided adding the test per @robertmaynard's https://github.com/NVIDIA/cudf-spark-jni/issues/4932#issuecomment-5179148812, but didn't think of putting it in `nvbench`…

PR #455, opened 2026-08-10
Reject conflicting stopping criterion options

### LINKED ISSUE #441: Global criterion option replay can silently discard user-provided stopping-criterion parameters
# Problem 

This is a follow-up/subissue of #408.

NVBench currently stores benchmark-scoped options that appear before any `--benchmark` as global argv fragments, then replays them onto each selected benchmark when that benchmark is created.

This replay model makes stopping-criterion options order-sensitive in a way that is hard to reason about. A later `--stopping-criterion` can reset criterion parameters that were explicitly set earlier, causing user input to be silently discarded.

# Examples

## Example 1: Input accepted, but `--min-time` is discarded

```
./bench --min-time 0.1 --stopping-criterion entropy --benchmark X
```

What happens today:

  1. `--min-time 0.1` is seen before any benchmark exists
  2. The current global stopping criterion is still the default, `stdrel`.
  3. `stdrel` accepts `--min-time`, so the option is stored globally.
  4. `--stopping-criterion entropy` is also stored globally.
  5. `--benchmark X` selects benchmark X.
  6. NVBench replays global args onto X in order:
     - apply `--min-time 0.1` to default `stdrel`
     - apply `--stopping-criterion entropy`
  7. Setting the criterion to `"entropy"` resets criterion params, so the earlier `--min-time 0.1` has no effect.

**Issue**: The command succeeds, but user input is ignored.

## Example 2: Reversed order correctly errors

```
./bench --stopping-criterion entropy --min-time 0.1 --benchmark X
```

What happens today:

  1. `--stopping-criterion entropy` is seen globally first.
  2. `--min-time 0.1` is then validated against parameters of  entropy criterion.
  3. `entropy` does not accept `--min-time`.
  4. NVBench fails with errors.

This behavior is better than that of Example 1.

**Issue**: Two commands that appear to express the same global intent behave differently based only on ordering.

## Example 3: Per-benchmark override of global defaults should remain valid

```
./bench --stopping-criterion entropy --min-r2 0.2 --benchmark X --stopping-criterion stdrel --min-time 0.7
```

What happens today:

 1. Global args configure `entropy` criterion with `--min-r2 0.2`.
 2. `--benchmark X` selects benchmark `X`.
 3. Global args are replayed, so `X` initially gets `entropy` criterion with `{"min-r2", 0.2}`
 4. Parsing continues after `--benchmark X`
 5. `--stopping-criterion stdrel` applies locally to `X`, replacing the inherited global criterion
 6. `--min-time 0.7` applies to local `stdrel`

This should remain valid: benchmark-local options should be allowed to override inherited global defaults because executable `./bench` may contain several benchmarks and the global setting of entropy criterion with parameters would apply to them.

## Example 4: Local benchmark options accepted, but --min-time is discarded

```
./bench --benchmark X --min-time 0.1 --stopping-criterion entropy
```

What happens today:

  1. `--benchmark X` selects benchmark `X`.
  2. `X` starts with the default stopping criterion, `stdrel`.
  3. `--min-time 0.1` is validated against `stdrel` and applied to `X`.
  4. `--stopping-criterion entropy` then applies to `X`.
  5. Setting the criterion to `entropy` resets criterion params, so the earlier local `--min-time 0.1` has no effect.

**Issue**: The command succeeds, but user input is ignored. 

This should error because a local criterion-specific parameter is discarded by a later local criterion change.

## Desired Behavior

NVBench should detect when a `--stopping-criterion` selection would discard stopping-criterion parameters explicitly configured earlier in the same option scope.

For example, these should each error:

```
./bench --min-time 0.1 --stopping-criterion entropy --benchmark X
```

```
./bench --benchmark X --min-time 0.1 --stopping-criterion entropy
```

Suggested error:

```
--stopping-criterion entropy would discard previously specified stopping-criterion parameter(s): --min-time.
Move --stopping-criterion before criterion-specific parameters.
```

But these inputs should remain valid because the local benchmark options intentionally override inherited global options:

```
./bench --stopping-criterion entropy --min-r2 0.2 --benchmark X --stopping-criterion stdrel --min-time 0.7
```

```
./bench --min-time 0.7 --benchmark Y --stopping-criterion entropy --min-r2 0.2
```

The inputs like `./bench --stopping-criterion entropy --stopping-criterion sample-count` and `./bench --benchmark X --stopping-criterion entropy --stopping-criterion sample-count` should also each generate an error.

```
--stopping-criterion sample-count would overwrite a stopping criterion already specified in this option scope: --stopping-criterion entropy.
Move benchmark-specific overrides after --benchmark, or remove the earlier --stopping-criterion.
```

## Implementation notes

A robust fix likely needs parser-side provenance tracking for stopping-criterion configuration. 

The key distinction is between configuration that is **inherited from global replay** and configuration that was **explicitly specified in the current option scope**.

This can be tracked with the following state:

```cpp
  struct criterion_cli_scope_state
  {
    std::optional<std::string> explicitly_selected_criterion;
    std::vector<std::string> explicitly_set_criterion_params;
  };
```

The parser would maintain:

 - one global `criterion_cli_scope_state` for options seen before any `--benchmark`
 - one local `criterion_cli_scope_state` for the currently active benchmark

When parsing an explicit `--stopping-criterion` in the current scope:

  - error if `explicitly_selected_criterion` is already set, because the new criterion would overwrite an earlier explicit criterion selection in the same scope 
  - error if `explicitly_set_criterion_params` is non-empty, because the new criterion would discard explicitly configured parameters in the same scope
  - otherwise record the selected criterion and apply it

When parsing an explicit stopping-criterion parameter in the current scope:

 - validate it against the current active criterion
 - record the parameter name in `explicitly_set_criterion_params`

During global replay onto a benchmark:

  - replayed global criterion settings should configure the benchmark
  - replayed global criterion settings should be treated as inherited, not locally explicit
  - after replay, the benchmark-local `criterion_cli_scope_state` should be empty

ahmedtaha100:
@oleksandr-pavlyk could you please vet this PR for copy-pr-bot when you have a chance? Thank you

Closes #441

## Summary

Reject `--stopping-criterion` changes that would discard parameters or overwrite a criterion set earlier in the same scope. Benchmark-local criterion options can still override inherited global settings.

## Testing

- `pre-commit run --files nvbench/option_parser.cuh nvbench/option_parser.cu testing/option_parser.cu`
- `ctest --test-dir /build --output-on-failure` (46/46 passed)

### DIFF
--- a/nvbench/option_parser.cu
+++ b/nvbench/option_parser.cu
@@ -133,6 +133,20 @@ std::string format_criterion_param_options(const nvbench::criterion_params &para
   return fmt::to_string(buffer);
 }
 
+std::string format_cli_options(const std::vector<std::string> &options)
+{
+  fmt::memory_buffer buffer;
+  for (std::size_t i = 0; i < options.size(); ++i)
+  {
+    if (i != 0)
+    {
+      fmt::format_to(fmt::appender(buffer), ", ");
+    }
+    fmt::format_to(fmt::appender(buffer), "{}", options[i]);
+  }
+  return fmt::to_string(buffer);
+}
+
 [[noreturn]] void
 throw_unrecognized_criterion_param(const std::string &prop_arg,
                                    const std::string &criterion_name,
@@ -441,6 +455,8 @@ void option_parser::parse(std::vector<std::string> args)
 void option_parser::parse_impl()
 {
   m_global_benchmark_args.clear();
+  m_global_criterion_cli_scope_state = {};
+  m_local_criterion_cli_scope_state  = {};
 
   // Initialize to all devices:
   m_recent_devices = nvbench::device_manager::get().get_devices();
@@ -845,16 +861,43 @@ catch (std::exception &e)
 void option_parser::set_stopping_criterion(const std::string &criterion)
 try
 {
+  auto &scope_state = m_benchmarks.empty() ? m_global_criterion_cli_scope_state
+                                           : m_local_criterion_cli_scope_state;
+
+  if (scope_state.explicitly_selected_criterion)
+  {
+    NVBENCH_THROW(std::runtime_error,
+                  "--stopping-criterion {} would overwrite a stopping criterion already "
+                  "specified in this option scope: --stopping-criterion {}.\n"
+                  "Move benchmark-specific overrides after --benchmark, or remove the earlier "
+                  "--stopping-criterion.",
+                  criterion,
+                  *scope_state.explicitly_selected_criterion);
+  }
+
+  if (!scope_state.explicitly_set_criterion_params.empty())
+  {
+    NVBENCH_THROW(std::runtime_error,
+                  "--stopping-criterion {} would discard previously specified stopping-criterion "
+                  "parameter(s): {}.\n"
+                  "Move --stopping-criterion before criterion-specific parameters.",
+                  criterion,
+                  format_cli_options(scope_state.explicitly_set_criterion_params));
+  }
+
   // If no active benchmark, save args as global.
   if (m_benchmarks.empty())
   {
     m_global_benchmark_args.push_back("--stopping-criterion");
     m_global_benchmark_args.push_back(criterion);
-    return;
+  }
+  else
+  {
+    benchmark_base &bench = *m_benchmarks.back();
+    bench.set_stopping_criterion(criterion);
   }
 
-  benchmark_base &bench = *m_benchmarks.back();
-  bench.set_stopping_criterion(criterion);
+  scope_state.explicitly_selected_criterion = criterion;
 }
 catch (std::exception &e)
 {
@@ -917,7 +960,9 @@ catch (std::exception &e)
 
 void option_parser::replay_global_args()
 {
+  m_local_criterion_cli_scope_state = {};
   this->parse_range(m_global_benchmark_args.cbegin(), m_global_benchmark_args.cend());
+  m_local_criterion_cli_scope_state = {};
 }
 
 void option_parser::update_devices(const std::string &devices)
@@ -1124,6 +1169,8 @@ void option_parser::update_criterion_prop(const std::string &prop_arg,
 try
 {
   const std::string name(prop_arg.begin() + 2, prop_arg.end());
+  auto &scope_state = m_benchmarks.empty() ? m_global_criterion_cli_scope_state
+                                           : m_local_criterion_cli_scope_state;
 
   // If no active benchmark, save args as global.
   if (m_benchmarks.empty())
@@ -1137,6 +1184,7 @@ try
 
     m_global_benchmark_args.push_back(prop_arg);
     m_global_benchmark_args.push_back(prop_val);
+    scope_state.explicitly_set_criterion_params.push_back(prop_arg);
     return;
   }
 
@@ -1174,6 +1222,8 @@ try
   {
     NVBENCH_THROW(std::runtime_error, "Unrecognized type for property: `{}`", name);
   }
+
+  scope_state.explicitly_set_criterion_params.push_back(prop_arg);
 }
 catch (std::exception &e)
 {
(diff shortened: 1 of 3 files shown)


oleksandr-pavlyk:
@coderabbitai full review

oleksandr-pavlyk:
The Python example failed because of this change (reported as #456). It is fixed in #457 by modifying the example.

Until that is merged, this PR remains blocked by CI checks.

nvbench/option_parser.cu
```diff
@@ -133,6 +133,20 @@ std::string format_criterion_param_options(const nvbench::criterion_params &para
...
+{
+  fmt::memory_buffer buffer;
+  for (std::size_t i = 0; i < options.size(); ++i)
+  {
+    if (i != 0)
+    {
+      fmt::format_to(fmt::appender(buffer), ", ");
+    }
+    fmt::format_to(fmt::appender(buffer), "{}", options[i]);
+  }
```
oleksandr-pavlyk: Suggestion: use iteration over container to improve clarity and remain consistent with the style used in `format_criterion_param_options`

```suggestion
  bool first = true;
  for (const auto &option: options)
  {
    if (!first)
    {
      fmt::format_to(fmt::appender(buffer), ", ");
    }
    fmt::format_to(fmt::appender(buffer), "{}", option);
    first = false;
  }
```

testing/option_parser.cu
```diff
@@ -1549,6 +1581,26 @@ void test_stopping_criterion()
...
+      "stdrel",
+      "--min-time",
+      "0.7",
+    });
+    const auto &states = parser_to_states(parser);
+
+    ASSERT(states.size() == 1);
+    ASSERT(states[0].get_stopping_criterion() == "stdrel");
+    ASSERT(states[0].get_criterion_params().get_float64("min-time") == 0.7);
+  }
```
oleksandr-pavlyk, line 1603: One additional regression test would be useful: verify local scope does not leak across benchmarks. 

Please add 

```diff
diff --git a/testing/option_parser.cu b/testing/option_parser.cu
index dfc186b..3a3f1cd 100644
--- a/testing/option_parser.cu
+++ b/testing/option_parser.cu
@@ -1601,6 +1601,38 @@ void test_stopping_criterion()
     ASSERT(states[0].get_stopping_criterion() == "stdrel");
     ASSERT(states[0].get_criterion_params().get_float64("min-time") == 0.7);
   }
+  { // Per-benchmark criterion state does not leak to the next benchmark:
+    nvbench::option_parser parser;
+    parser.parse({
+      "--stopping-criterion",
+      "entropy",
+      "--min-r2",
+      "0.2",
+      "--benchmark",
+      "DummyBench",
+      "--stopping-criterion",
+      "stdrel",
+      "--min-time",
+      "0.7",
+      "--benchmark",
+      "TestBench",
+    });
+
+    auto &benches = parser.get_benchmarks();
+    ASSERT(benches.size() == 2);
+
+    benches[0]->run();
+    const auto &dummy_states = benches[0]->get_states();
+    ASSERT(dummy_states.size() == 1);
+    ASSERT(dummy_states[0].get_stopping_criterion() == "stdrel");
+    ASSERT(dummy_states[0].get_criterion_params().get_float64("min-time") == 0.7);
+
+    benches[1]->run();
+    const auto &test_states = benches[1]->get_states();
+    ASSERT(test_states.size() == 9);
+    ASSERT(test_states[0].get_stopping_criterion() == "entropy");
+    ASSERT(test_states[0].get_criterion_params().get_float64("min-r2") == 0.2);
+  }
   { // Sample-count criterion default params
     nvbench::option_parser parser;
     parser.parse({
```

nvbench/option_parser.cu
ahmedtaha100: Updated this to use a range based loop. Thank You!

testing/option_parser.cu
ahmedtaha100: Added the regression test. Thank you!

oleksandr-pavlyk (approved):

